PR #2, opened 2019-11-20
Feature - SDK

markogresak:
https://app.clubhouse.io/prezly/story/10308/extract-sdk-functionality

There's a ton of changes because of the copied types. These files are exactly the same as in the backend app. The only change is the `Entity` type, which is generic and accepts optional `id` type, while still defaulting to `number` (to match the original).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,8 @@ build/Release
 
 node_modules/
 
+/coverage
+
 # TypeScript cache
 *.tsbuildinfo
 
--- a/src/Api/Api.ts
+++ b/src/Api/Api.ts
@@ -1,13 +1,13 @@
 import { Method, HeadersMap, Response } from './types';
 import createRequest from './createRequest';
 
-const withAccessToken = (headers?: HeadersMap, accessToken?: string): HeadersMap => ({
+const withAccessToken = (headers: HeadersMap = {}, accessToken: string): HeadersMap => ({
     ...headers,
-    ...(accessToken ? { 'x-access-token': accessToken } : {}),
+    'x-access-token': accessToken,
 });
 
 interface Params {
-    accessToken?: string;
+    accessToken: string;
     headers?: HeadersMap;
     query?: object;
 }
@@ -17,10 +17,7 @@ interface ParamsWithPayload extends Params {
 }
 
 const Api = {
-    get: <P = any>(
-        url: string,
-        { accessToken, headers, query }: Params = {},
-    ): Promise<Response<P>> =>
+    get: <P = any>(url: string, { accessToken, headers, query }: Params): Promise<Response<P>> =>
         createRequest(url, {
             headers: withAccessToken(headers, accessToken),
             method: Method.GET,
@@ -29,7 +26,7 @@ const Api = {
 
     post: <P = any>(
         url: string,
-        { accessToken, headers, payload, query }: ParamsWithPayload = {},
+        { accessToken, headers, payload, query }: ParamsWithPayload,
     ): Promise<Response<P>> =>
         createRequest(url, {
             headers: withAccessToken(headers, accessToken),
@@ -40,7 +37,7 @@ const Api = {
 
     put: <P = any>(
         url: string,
-        { accessToken, headers, payload, query }: ParamsWithPayload = {},
+        { accessToken, headers, payload, query }: ParamsWithPayload,
     ): Promise<Response<P>> =>
         createRequest(url, {
             headers: withAccessToken(headers, accessToken),
@@ -51,7 +48,7 @@ const Api = {
 
     patch: <P = any>(
         url: string,
-        { accessToken, headers, payload, query }: ParamsWithPayload = {},
+        { accessToken, headers, payload, query }: ParamsWithPayload,
     ): Promise<Response<P>> =>
         createRequest(url, {
             headers: withAccessToken(headers, accessToken),
@@ -62,7 +59,7 @@ const Api = {
 
     delete: <P = any>(
         url: string,
-        { accessToken, headers, payload, query }: ParamsWithPayload = {},
+        { accessToken, headers, payload, query }: ParamsWithPayload,
     ): Promise<Response<P>> =>
         createRequest(url, {
             headers: withAccessToken(headers, accessToken),
--- a/src/Api/Api.ts
+++ b/src/Api/Api.ts
@@ -1,75 +1,74 @@
 import { Method, HeadersMap, Response } from './types';
 import createRequest from './createRequest';
 
+const withAccessToken = (headers?: HeadersMap, accessToken?: string): HeadersMap => ({
+    ...headers,
+    ...(accessToken ? { 'x-access-token': accessToken } : {}),
+});
+
+interface Params {
+    accessToken?: string;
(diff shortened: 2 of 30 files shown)


src/Api/Api.ts
```diff
@@ -1,75 +1,74 @@
 import { Method, HeadersMap, Response } from './types';
 import createRequest from './createRequest';
 
+const withAccessToken = (headers?: HeadersMap, accessToken?: string): HeadersMap => ({
+    ...headers,
+    ...(accessToken ? { 'x-access-token': accessToken } : {}),
+});
+
+interface Params {
+    accessToken?: string;
```
kamilmielnik: Should it not be required?

```suggestion
    accessToken: string;
```

src/Sdk/Coverage/types.ts
```diff
@@ -0,0 +1,12 @@
+import { Coverage, Pagination } from 'types';
+
+export interface Options {
+    token: string;
+    baseUrl?: string;
```
kamilmielnik: I see that later you use `Required<Options>` in Coverages constructor. 
Why is `baseUrl` optional here? I see that the usage in Coverages constructor is the only one, so we can simplify things.

```suggestion
    baseUrl: string;
```

And then remove `Required<>` from Coverages constructor.

src/Sdk/routing.ts
```diff
@@ -0,0 +1,3 @@
+export const BASE_URL = 'https://api.prezly.com';
+
+export const coverageUrl = '/v2/coverage';
```
kamilmielnik: Seems inconsistent
```suggestion
export const COVERAGE_URL = '/v2/coverage';
```

src/types/UserRef.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * @see sfGuardUser::toApiReferenceArray()
```
kamilmielnik: I don't think this comment belongs in the SDK as it refers to something in a private repository.

src/types/UserAccountRef.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * @see UserLicense::toApiArray()
```
kamilmielnik: I don't think this comment belongs in the SDK as it refers to something in a private repository.

src/types/EntitiesMap.ts
```diff
@@ -0,0 +1,7 @@
+export interface Entity<T extends number | string = number> {
+    id: T;
+}
+
+export interface EntitiesMap<E = Entity> {
+    [index: number]: E;
```
kamilmielnik: This type is wrong now. `index` can be a `string` as well (it's always casted to string anyway by the nature of what object keys are in JS, hmm). So the type is inaccurate, and the name as well (`key` suits better if it can also be a `string`).

It should be something like this:

```suggestion
    [key: T]: E<T>;
```

Or we can just remove it since `EntitiesMap` is not used anywhere in this SDK yet.

src/Sdk/Coverage/Coverage.ts
```diff
@@ -0,0 +1,51 @@
+import { Api } from 'Api';
+import { Coverage } from 'types';
+
+import { Options } from '../types';
+import { coverageUrl } from '../routing';
+import { getItemId } from '../utils';
+
+import { CoverageListResponse } from './types';
+
+export default class Coverages {
```
e1himself: FYI `Coverage` does not have plural form. It's just `Coverage`.

src/types/EntitiesMap.ts
e1himself: > It should be something like this: 
>
> ```
> [key: T]: E<T>;
> ```

FYI TS will not allow this.

.gitignore
```diff
@@ -25,6 +25,8 @@ build/Release
 
 node_modules/
 
+/coverage
```
kudlajz: What is this folder for?

src/Sdk/Coverage/types.ts
markogresak: It's also used in the `Sdk`, where it should be optional.

src/types/EntitiesMap.ts
kudlajz: Since it's not used anywhere, let's just remove it. Furthermore, most of the entities collections are arrays anyway, right? So, we don't even need this interface.

e1himself: Lukas is right, we don't need it in this repo. So feel free to remove it! :+1:

e1himself: Also there's no real need for `Entity` type *in this repo* as well.

src/Sdk/routing.ts
markogresak: I did this deliberatley to match our `appRouting`. 

The decision to make `BASE_URL` constant case is because I see it as a different type. I can move it to `Sdk.ts` because it's only used there as the default value.

src/Sdk/Coverage/types.ts
kamilmielnik: > It's also used in the Sdk, where it should be optional.

Sdk uses `Options` interface from `src/Sdk/types.ts`, so a different one. 

So my suggestion is still valid.

markogresak: Nope, it imports `import { Options } from '../types';` So the type here is unused in the version of the code when you commented. 

So this one is unnecessarily duplicated. I moved the type to the `Coverage.ts` file, similarly to how we have `Props` next to the component. And I have applied the your suggestion to not use the `Required`.

src/Sdk/Coverage/Coverage.ts
markogresak: Oh, right. But when I was trying to rename it, there is a conflict with the `Coverage` data type. I've picked `CoverageSdk` to avoid this conflict. Or does anyone have a better naming suggestion?

.gitignore
markogresak: Coverages from `jest`. We do not use it, but someone might want to run `jest --coverage`. For example, I did that. So this is more "just in case" .  The `/coverage` is used to match only the folder in project root, otherwise it also matches the `Coverage` inside `Sdk` folder.

src/Api/Api.ts
markogresak: 78fb90580c962f9d001fc254714f6408ef6ba61a

src/types/EntitiesMap.ts
markogresak: Removed `EntitesMap` 39a08864cbb930cdf0a4724e41b3a5f1a8f7a775

src/types/UserAccountRef.ts
markogresak: Removed be1e4042d63d74a5d77e5c81b2361cd3cb652ac2

src/types/UserRef.ts
markogresak: Removed be1e4042d63d74a5d77e5c81b2361cd3cb652ac2

src/Sdk/routing.ts
markogresak: Changed 0c85b29578e2fa7434fbff595688c14111942abb

PR #3, opened 2019-11-20
Chore - Publish

markogresak:
https://app.clubhouse.io/prezly/story/10366/publish-the-package

This is the difference of #2 to publish a working version:
 - fix repository url
 - add [`np`](https://www.npmjs.com/package/np) and config
 - fixed import paths because non-relative imports did not work with the built script

Also switched from `yarn` (`yarn.lock`) to `npm` (`package-lock.json`) as we discussed. We can go back to `yarn` if we decide it's suitable.

This is already publised as a [pre-release](https://github.com/prezly/javascript-sdk/releases/tag/v1.0.0-2) and installable under `@prezly/sdk@beta`. A full release is planned once we implement all SDK functions.


package.json
```diff
@@ -1,36 +1,43 @@
 {
     "name": "@prezly/sdk",
-    "version": "1.0.0",
+    "version": "1.0.0-2",
```
kamilmielnik: That's exactly why I wrote the following in the story :wink: 

> Publish as `@prezly/sdk` - start with a `1.0.0-beta.0`/`1.0.0-rc.0` version, publish it, test it as a dependency in another project, and once you're sure it works, you can publish it as `1.0.0`

package.json
```diff
@@ -1,36 +1,43 @@
...
     "description": "Prezly API SDK",
     "main": "dist/index.js",
+    "files": [
```
kamilmielnik: So I'm trying to use this package (yes, `"@prezly/sdk": "^1.0.0-2"`), I have this code:

```ts
import PrezlySdk from '@prezly/sdk';

const prezlySdk = new PrezlySdk({ accessToken: PREZLY_ACCESS_TOKEN });
```
and I get this error:

```bash
src/index.ts:13:27 - error TS2351: This expression is not constructable.
  Type 'typeof import("/home/kamil/projects/prezly/kantar-sync/node_modules/@prezly/sdk/dist/index")' has no construct signatures.

13     const prezlySdk = new PrezlySdk({ accessToken: PREZLY_ACCESS_TOKEN });
```

`node_modules/@prezly/sdk/dist/index.d.ts` contains only this:
```ts
export {};
```

src/index.ts
```diff
@@ -1 +1,3 @@
-export { Sdk as default } from 'Sdk';
+import { default as PrezlySdk } from './Sdk';
+
+module.exports = PrezlySdk;
```
e1himself: Why not just Prezly?

package.json
e1himself: Another trick I use -- is to install `master` directly and test the package before releasing a version.

markogresak: @e1himself you could use [`npm link`](https://docs.npmjs.com/cli/link)

It's exactly the same and doesn't require you to have git.

1. Run `npm link` in source project, in this case it's `prezly/javascript-sdk`
2. Run `npm link [package]`, in this case `npm link @prezly/sdk`

The first will regirster a global link on your device. The second will install from your local device based on the link. It's symlink taken to the level of npm packages.

markogresak: > That's exactly why I wrote the following in the story

@kamilmielnik I don't understand, is it because it took me 3 versions to get it right? 😄Or did you have something else in mind?

kamilmielnik: Well, yeah :smile: 

I hoped that when this story is complete the version will be `1.0.0`, not `1.0.0-2` . 
But :man_shrugging: whatever :smile:

package.json
markogresak: Hmm. I was assuming commonjs, i.e. `require(@prezly/sdk)`. I assume the `import` expect the 'module.default' key, which doesn't work. I will change it with the next release.

package.json
markogresak: @kamilmielnik yes, it will be. This is the "premajor" release. 

image

package.json
markogresak: Should be fixed with 8c7c07154f6b1277c07432cd90eb643e628b9473.

src/index.ts
markogresak: Changed the export with 8c7c07154f6b1277c07432cd90eb643e628b9473.

PR #4, opened 2019-11-21
Feature - More SDK functions

markogresak:
https://app.clubhouse.io/prezly/story/10308/extract-sdk-functionality

PR #6, opened 2019-11-21
Feature - More SDK Functions

markogresak:
https://app.clubhouse.io/prezly/story/10308/extract-sdk-functionality

#4 showed incorrect diff because this branch is based off #3 and the commits were merged in the incorrect branch. I closed #4 in favor of this PR.  

Changes in this PR are already published as `@prezly/sdk@beta`.

### DIFF
--- a/src/Sdk/Coverage/Coverage.ts
+++ b/src/Sdk/Coverage/Coverage.ts
@@ -4,7 +4,10 @@ import { Coverage } from '../../types';
 import { coverageUrl } from '../routing';
 import { getItemId } from '../utils';
 
-import { CoverageListResponse } from './types';
+import {
+    CoverageListResponse,
+    CoverageSearchOptions,
+} from './types';
 
 interface Options {
     accessToken: string;
@@ -20,9 +23,16 @@ export default class CoverageSdk {
         this.url = `${baseUrl}/${coverageUrl}`;
     }
 
-    async list(): Promise<CoverageListResponse> {
+    async list(options: CoverageSearchOptions = {}): Promise<CoverageListResponse> {
+        const { jsonQuery, page, pageSize, sortOrder } = options;
         const response = await Api.get<CoverageListResponse>(this.url, {
             accessToken: this.accessToken,
+            query: {
+                page,
+                pageSize,
+                query: jsonQuery,
+                sort: sortOrder,
+            },
         });
         return response.payload;
     }
--- a/src/Sdk/Coverage/Coverage.ts
+++ b/src/Sdk/Coverage/Coverage.ts
@@ -44,6 +44,12 @@ export default class CoverageSdk {
         return response.payload.coverage;
     }
 
+    async getByExternalReferenceId(externalReferenceId: string): Promise<Coverage | null> {
+        const jsonQuery = JSON.stringify({ external_reference_id: { $in: [externalReferenceId] } });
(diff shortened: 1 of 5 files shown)


src/Sdk/Coverage/Coverage.ts
```diff
@@ -34,15 +46,32 @@ export default class CoverageSdk {
...
-        throw new Error('Not implemented!');
+    async getByExternalReferenceId(externalReferenceId: string): Promise<Coverage | null> {
+        const jsonQuery = JSON.stringify({ external_reference_id: { $in: [externalReferenceId] } });
+        const { coverage } = await this.list({ jsonQuery });
+        return coverage[0] || null;
     }
 
-    async update(): Promise<Coverage> {
-        throw new Error('Not implemented!');
+    async create(payload: CoverageCreateRequest): Promise<Coverage> {
```
kamilmielnik: I'm trying the API out and I have an interesting case.

I have passed an object that matches `CoverageCreateRequest` interface as payload, but it had some extra attributes. API failed with HTTP 422 Unprocessable Entity.

But I feel like TypeScript should have prevented me from doing this. So we could improve here by making sure no extra attributes are passed.

Is there some mechanism in TypeScript to disallow extra attributes?

markogresak: Hmm, it should work like that. TypeScript will not allow you to construct an object with properties that are not specified in the interface. You can see this with the `Props` type in our React code.

I think the problem is that TypeScript doesn't read the type and it assumes `create(...arguments: any[]): any`, so it allows you to pass in invalid `payload`.

kudlajz: I think you're not correct Marko. Interface only checks the declared properties, but any extra properties are not checked.

You can see it here: https://www.typescriptlang.org/play/#code/JYOwLgpgTgZghgYwgAgGIHt3IN4FgBQyRycAXMiAK4C2ARtANwHHK3lV2MEC+BBMlEAjDB0IZABN0GdAAoYmcjICUOZsR598CMQGcwJZAF4cJcgEYANK3IAmZNyba9BsJgCylBAAtjpsshWNsi21gjkAMwOTgRSMrJwyk5xmLJu6J4+SUA

I have found a blog talking about this issue and how to solve it here: https://fettblog.eu/typescript-match-the-exact-object-shape/

kamilmielnik: Nice find @kudlajz , that's exactly what I'd like to see here.

kamilmielnik: I was thinking on this over the weekend and I guess doing it in TS is not a solution for this, but a nice-to-have, at most.

Because if someone uses this lib in pure JS, they would still have the problem I initially reported (HTTP 422).

So I guess what we should do is something like this: 

```ts
const response = await Api.post<{ coverage: Coverage }>(this.url, {
    accessToken: this.accessToken,
    payload: {
        id: payload.id, // etc.
        /* map all `payload` attributes here */
    },
});
```

Because (given that SDK is provided with all information necessary to perform the API request), it's SDK's responsibility to make sure the API request is not malformed. 

Thoughts? @kudlajz @markogresak

kudlajz: I agree Kamil. It makes sense, especially for non-TS people and making it work this way, we don't need to do anything fancy with our interfaces either.

PR #7, opened 2019-11-21
Feature - replace node-fetch with cross-fetch

markogresak:
https://app.clubhouse.io/prezly/story/10498/support-both-node-and-browser

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,6 @@
         "typescript": "^3.7.2"
     },
     "dependencies": {
-        "cross-fetch": "^3.0.4",
         "query-string": "^6.9.0"
     },
     "scripts": {
--- a/package.json
+++ b/package.json
@@ -16,7 +16,6 @@
         "@babel/core": "^7.7.2",
         "@babel/preset-env": "^7.7.1",
         "@types/node": "^12.12.9",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -27,7 +26,7 @@
         "typescript": "^3.7.2"
     },
     "dependencies": {
-        "node-fetch": "^2.6.0",
+        "cross-fetch": "^3.0.4",
```
markogresak: The suggestion was to use `isomorphic-fetch`. But the last commit on the project was on 22 May 2016 (over 3 years ago!) so I consider it unmaintined. 
I have decided to pick `cross-fetch`, which does the same thing, but it's maintined better.

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,26 @@
+name: Node CI
```
markogresak: This is from master (https://github.com/prezly/javascript-sdk/commit/1182f85307a1b2863b0948b69c8bdfb5b1ecf474), but missing in the target branch.

It's automatically generated by the Github Actions interface. Runs `npm run build` and `npm test` in node@8, node@10, and node@12 (current LTS vesions).

PR #8, opened 2019-11-21
Feature - User Agent

markogresak:
https://app.clubhouse.io/prezly/story/10462/add-user-agent-header


src/Api/Api.ts
```diff
@@ -1,13 +1,9 @@
-import { Method, HeadersMap, ApiResponse } from './types';
-import createRequest from './createRequest';
+import { HeadersMap } from '../types';
 
-const withAccessToken = (headers: HeadersMap = {}, accessToken: string): HeadersMap => ({
-    ...headers,
-    'x-access-token': accessToken,
```
markogresak: Decided not to couple the `accessToken` with `Api`. The `BaseSdk` now takes care of that, along with the `User-Agent` header.

src/Sdk/Coverage/types.ts
```diff
@@ -14,9 +14,7 @@ export interface CoverageUpdateRequest {
     url?: string | null;
 }
 
-export interface CoverageCreateRequest extends CoverageUpdateRequest {
-    author: Contact['id'] | string;
-}
+export interface CoverageCreateRequest extends CoverageUpdateRequest {}
```
markogresak: Hijacked this PR to revert the incorrect type fix https://github.com/prezly/javascript-sdk/commit/0e66a7e6774566a76b9ca3da81302f6816a28e4f.
Mentioned by @e1himself here https://prezly.slack.com/archives/C029LTE0M/p1574349891215900?thread_ts=1574344190.213000&cid=C029LTE0M

src/Api/Api.ts
e1himself: May I ask why? 

AFAIK all Clean Code 101 articles are stating that composition should always be preferred over inheritance. If possible.

In my arrogant opinion 😜, headers and base URI are http concepts and it totally makes sense to couple them to http client instance.

kamilmielnik: > AFAIK all Clean Code 101 articles are stating that composition should always be preferred over inheritance. If possible.

I don't know what does composition/inheritance have to do with it...

> totally makes sense to couple them to http client instance.

But I agree :+1:

src/Sdk/BaseSdk.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/**
+ * Remove heading and trailing slashes.
+ * Examples:
+ *  - https://example.com/ -> https://example.com
+ *  - /v2/path -> v2/path
+ *  - /v2/path/ -> v2/path
+ */
+const sanitizeUrl = (url: string): string => url.replace(/^\/|\/$/g, '');
```
kamilmielnik: How we could change the code so that:
- somewhat complex regular expression is broken into 2 simpler regular expressions (no `|`, no `g`)
- code is expressive enough so it does not need comments
- we're left with less lines than initially

```suggestion
const removeLeadingSlash = (url: string) => url.replace(/^\//, '');
const removeTrailingSlash = (url: string) => url.replace(/\/$/, '');
const sanitizeUrl = (url: string) => removeTrailingSlash(removeLeadingSlash(url));
```

e1himself: Or we can keep the all-in-one regex, but rename the function to `trimSlashes()` -- so it's obvious what it does.

src/Api/Api.ts
markogresak: > composition should always be preferred over inheritance

I  do not follow. What inheritance? 
Why should `Api` module know about our specific implementation of `accessToken` header (which is not following the usual [`Authorization` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) pattern)? 
My idea of `Api` is to one day extract it and use as a dependency to our projects instead of copying it around. It should be composable enough that anyone else can use it in their code. For this reason, I don't want it to be opinionated of the `accessToken`. This change achieves my goal. It should be able to perform requests to *any* endpoint, not only to the Prezly API. 

On the other hand, the `Sdk` module is tied to our API, so it should be the part of the code responsible to provide correct authorization to the API. For the same reason, I've decided to place `User-Agent` header inside `Sdk` module.

e1himself: Me, Marko and Kamil had a call about this and resolved the confusion around this thread.

```js
const apiClient = new ApiClient({
    headers: {
        'X-Access-Token': token,
        'User-Agent': userAgent,
    },
    baseUri,
});

const coverage = new Coverage(apiClient);
```

markogresak: Changed with https://github.com/prezly/javascript-sdk/pull/8/commits/e439047640cf24dd1c2216c0212ad26284d7594d.

src/Sdk/Coverage/types.ts
markogresak: The type is now updated to match the latest type in the backend app (7678d4259f7025bff2b49e9a99597588a52615c9).

src/Sdk/ApiClient.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public buildUrl(sdkUrl: string): string {
+        return `${stripSlashes(this.baseUrl)}/${stripSlashes(sdkUrl)}`;
+    }
+
+    public buildUrlWithId(sdkUrl: string, itemOrItemId: string | Entity<number | string>): string {
+        const itemId = typeof itemOrItemId === 'object' ? itemOrItemId.id : itemOrItemId;
+        return `${this.buildUrl(sdkUrl)}/${itemId}`;
+    }
```
e1himself: I think this function does not belong here.

PR #9, opened 2019-11-21
Feature - SDK tests

markogresak:
No task for this. I've skipped doing any tests in the initial implementation of SDK (#2, #6), but I believe we should have some tests.


src/Sdk/Sdk.test.js
```diff
@@ -0,0 +1,209 @@
...
+    body: undefined,
+    headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json;charset=utf-8',
+        'User-Agent': DEFAULT_USER_AGENT,
+        'x-access-token': ACCESS_TOKEN,
+    },
+};
+
+function successJsonResponse(body, status = 200) {
```
markogresak: This is copied from `Api.test.js`. I wasn't sure if it was worth it to extract this into "test utils"  file.

PR #10, opened 2019-11-25
Fix - Move buildUrl and buildUrlWithId to SDK utils

markogresak:
Mentioned here: https://github.com/prezly/javascript-sdk/pull/8#discussion_r350152859

### DIFF
--- a/src/Sdk/ApiClient.ts
+++ b/src/Sdk/ApiClient.ts
@@ -1,6 +1,7 @@
 import { Api, ApiResponse, HeadersMap, Params, ParamsWithPayload } from '../Api';
 
 import { DEFAULT_USER_AGENT } from './constants';
+import { stripSlashes } from './utils';
 
 export interface Options {
     accessToken: string;
@@ -9,7 +10,7 @@ export interface Options {
 }
 
 export default class ApiClient {
-    public readonly baseUrl: string;
+    private readonly baseUrl: string;
     private readonly headers: HeadersMap;
 
     constructor({ accessToken, baseUrl, headers }: Options) {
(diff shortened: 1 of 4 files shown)


src/Sdk/Coverage/Coverage.ts
```diff
@@ -21,7 +22,7 @@ export default class CoverageSdk {
     async list(options: CoverageSearchOptions = {}): Promise<CoverageListResponse> {
         const { jsonQuery, page, pageSize, sortOrder } = options;
         const response = await this.apiClient.get<CoverageListResponse>(
-            this.apiClient.buildUrl(this.sdkUrl),
+            buildUrl(this.apiClient.baseUrl, this.sdkUrl),
```
e1himself: Would it make sense if `ApiClient` were operating over relative URLs? Then you don't need to expose this detail at all. Something like this:

```ts
const { coverage } = await apiClient.get(`/v2/coverage/${coverageId}`);
```

e1himself: > `this.sdkUrl`

There is an advantage of using `appRouting` object directly without destructureing it -- you can trace the usage of properties down to every API call. But we lose that once a value is destructured and reassigned. 

Is it worth to introduce all the intermediate properties and instance state when we can just use `routing.coverageUrl`?

markogresak: Good suggestion. Changed with fb42885fc15c8f794ef359fc7d1cdf96c7050146.

markogresak: Yes, looks better this way and it's easier to track usages as you've pointed out. Done with 69de8e1706387578b8d6dd63557c8fd60e4fb194

src/Sdk/ApiClient.ts
```diff
@@ -105,4 +90,8 @@ export default class ApiClient {
             query,
         });
     }
+
+    private buildEndpointUrl(endpointUri: string): string {
+        return `${stripSlashes(this.baseUrl)}/${stripSlashes(endpointUri)}`;
```
markogresak: I'm doing this because I don't want to make any assumtions of the `baseUrl` since it can be a value from outside the module.
I think it's easier and better DX if we sanitize the url to how we want them instead of explaining what the value should be and leaving options for a small mistake like a trailing slash.

PR #11, opened 2020-02-28
Feature  - Coverage API updates

e1himself:
- Fix a bug with `pageSize` not being passed as `limit` request param

- Fix wrong coverage ID type (it is actually `number`, but `string` was declared before)

- Update Coverage interface with the latest changes in Prezly APIs (`url` and `published_at` properties)

- Add support for `headline` property
     https://app.clubhouse.io/prezly/story/11641/update-prezly-sdk-package
  
   - Add new `headline` property to Coverage interface
   - Add new `headline?` property to Coverage create/update requests

   <br>

- Add support for soft-deleted coverage
   https://app.clubhouse.io/prezly/story/13255/update-prezly-javascript-sdk

   - Add new `is_deleted` property to Coverage interface
   - Add new `include_deleted` param to Coverage List API function
   - Always use `include_deleted: true` for `external_reference_id` lookups

### DIFF
--- a/src/Sdk/Sdk.test.js
+++ b/src/Sdk/Sdk.test.js
@@ -90,7 +90,7 @@ describe('Sdk', () => {
             });
             const result = await prezlySdk.coverage.getByExternalReferenceId(externalReferenceId);
 
-            const search = `?query=%7B%22external_reference_id%22%3A%7B%22%24in%22%3A%5B%22${externalReferenceId}%22%5D%7D%7D`;
+            const search = `?include_deleted=on&query=%7B%22external_reference_id%22%3A%7B%22%24in%22%3A%5B%22${externalReferenceId}%22%5D%7D%7D`;
             expect(fetch).toHaveBeenCalledWith(`${defaultCoverageApiUrl}${search}`, {
                 ...DEFAULT_REQUEST_PROPS,
                 method: Method.GET,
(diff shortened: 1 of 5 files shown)


src/Sdk/Sdk.test.js
```diff
@@ -90,7 +90,7 @@ describe('Sdk', () => {
             });
             const result = await prezlySdk.coverage.getByExternalReferenceId(externalReferenceId);
 
-            const search = `?query=%7B%22external_reference_id%22%3A%7B%22%24in%22%3A%5B%22${externalReferenceId}%22%5D%7D%7D`;
+            const search = `?include_deleted=on&query=%7B%22external_reference_id%22%3A%7B%22%24in%22%3A%5B%22${externalReferenceId}%22%5D%7D%7D`;
```
kamilmielnik: A side note: we're sort of lucky that these tests pass.
```
?query=%7B%22external_reference_id%22%3A%7B%22%24in%22%3A%5B%22${externalReferenceId}%22%5D%7D%7D&include_deleted=on
```

would also be a valid value. Running this test in a different environment could theoretically fail. But we can skip it for now though.

e1himself:
@kamilmielnik do you think this update should be released as a major version bump?

kamilmielnik:
> @kamilmielnik do you think this update should be released as a major version bump?

@e1himself since the type of `id ` has changed, it's not backwards compatible, so yes: major version bump makes sense.

PR #13, opened 2020-02-28
Feature - Export all types & request/response interfaces

e1himself:
It's nice to have access to these interfaces when writing consuming libraries.

Like Alex has it here for Belga syncing code:
https://github.com/prezly/belga-sync/blob/247774b69ba60d7380af62937b3cfbd3a1b6b875/src/util/belga-importer.ts#L5

```ts
import { CoverageCreateRequest } from '@prezly/sdk/dist/Sdk/Coverage/types';
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,1 +1,18 @@
-export { default } from './Sdk';
+export { default } from './Sdk/Sdk';
+export {
+    Contact,
+    Coverage,
+    Entity,
+    Pagination,
+    RoomRef,
+    Story,
+} from './types'
+export {
+    Options as ClientOptions,
+} from './Sdk/types'
+export {
+    CoverageCreateRequest,
+    CoverageListResponse,
+    CoverageSearchOptions,
+    CoverageUpdateRequest,
+} from './Sdk/Coverage/types'
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,11 @@
-export { default } from './Sdk/Sdk';
+export {
+    default,
+    ClientOptions,
+    CoverageCreateRequest,
+    CoverageListResponse,
+    CoverageSearchOptions,
+    CoverageUpdateRequest,
+} from './Sdk';
 export {
     Contact,
     Coverage,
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -1 +1,18 @@
-export { default } from './Sdk';
+export { default } from './Sdk/Sdk';
```
kudlajz: I think this can stay
```suggestion
export { default } from './Sdk';
```

no?

src/index.ts
```diff
@@ -1 +1,18 @@
...
+} from './types'
+export {
+    Options as ClientOptions,
+} from './Sdk/types'
+export {
+    CoverageCreateRequest,
+    CoverageListResponse,
+    CoverageSearchOptions,
+    CoverageUpdateRequest,
+} from './Sdk/Coverage/types'
```
kamilmielnik: I don't think we should go this "deep" - that's not what we do in other repos.

I'd expect `Sdk/Coverage/index.ts` to re-export all these from `./types`, and then `Sdk/index.ts` to re-export all these from `./Coverage`.

I imagine this (`src/index/.ts`) file like this:
```ts
export {
    default,
    CoverageCreateRequest,
    CoverageListResponse,
    CoverageSearchOptions,
    CoverageUpdateRequest,
    Options as ClientOptions,
} from './Sdk';
export {
    Contact,
    Coverage,
    Entity,
    Pagination,
    RoomRef,
    Story,
} from './types';
```

src/index.ts
```diff
@@ -1 +1,18 @@
-export { default } from './Sdk';
+export { default } from './Sdk/Sdk';
+export {
+    Contact,
+    Coverage,
+    Entity,
+    Pagination,
+    RoomRef,
+    Story,
+} from './types'
```
kamilmielnik:
```suggestion
} from './types';
```

The build does not fail - it looks like we're not enforcing `prettier` in this project in any way, even though we have `.prettierrc` and appropriate script in `package.json`.

e1himself:
Thanks for your reviews guys! :+1:

PR #14, opened 2020-03-02
Feature - Include oembed as a property of coverage

atrauzzi:
[[branch CH13066]](https://app.clubhouse.io/prezly/story/13066/implement-a-nodejs-cron-job-to-feed-belga-api-results-into-prezly)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@prezly/sdk",
-    "version": "2.1.0",
+    "version": "2.1.1",
     "description": "Prezly API SDK",
     "main": "dist/index.js",
     "files": [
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@prezly/sdk",
-    "version": "2.1.0",
+    "version": "2.1.1",
```
e1himself: For the record: this is done automatically by `np` command.

All you had to do is to run `np patch` -- it'd do the rest, including bumping this version and publishing the package to `npm`.

atrauzzi: I would looooooove to [automate this](https://discourse.prezly.io/t/versioning-in-ci/160). 🙂

e1himself: File a PR :)

PR #18, opened 2020-12-28
Feature - Stories and Newsrooms SDKs

e1himself:
- Implement Stories SDK
- Implement Newsrooms SDK
- I had to reorganize some types here and there to make it more convenient to maintain.


src/Sdk/Coverage/types.ts
```diff
@@ -7,7 +7,7 @@ export interface CoverageUpdateRequest {
     author?: Contact['id'] | string | null;
     external_reference_id?: string;
     headline?: string;
-    newsroom?: RoomRef['id'] | null;
+    newsroom?: NewsroomRef['id'] | null;
```
e1himself, line 10: This was updated in the API, but this interface here was left outdated.

src/types/Story.ts
```diff
@@ -44,7 +82,7 @@ export default interface Story extends Entity {
         translate: string | null;
     };
 
-    room: RoomRef;
```
e1himself, line 47: This was updated in the API, but this interface here was left outdated.

markogresak (approved):

PR #21, opened 2021-01-20
[DEV-1327] Feature - Newsroom Contacts SDK

e1himself:
DEV-1327 

- Implement Newsroom Contacts SDK
- Organize secondary models types into a sub-folder
- Export all in-depth SDK types


src/types/index.ts
```diff
@@ -13,10 +16,14 @@ export {
...
-export { default as OEmbedInfo } from './OEmbedInfo';
-export { default as Pagination } from './Pagination';
+export {
+    UploadcareFile,
+    UploadcareImage,
+    OEmbedInfo,
+    Pagination,
+    SelectionMode,
+    SelectionValue,
+} from './secondary';
```
kamilmielnik, line 26: I never saw this pattern before.
What is a "secondary" entity?
Why does that distinction help with?

src/Sdk/NewsroomContacts/types.ts
```diff
@@ -0,0 +1,36 @@
...
+    search?: string;
+    /**
+     * Prezly Query Language expression to filter contacts by.
+     */
+    query?: string;
+}
+
+export interface NewsroomContactCreateRequest {
+    name: string;
+    email?: string | null;
```
kamilmielnik, line 23: I thought [our agreements](https://discourse.prezly.io/t/ordering-properties-alphabetically-in-javascript/186/9?u=kamil) are indefinite and Prezly-wide.

It does not have to be done in this PR, I just want us to:
- be consistent
- abide and respect team decisions

Of course we can re-discuss any decision, I just think in this case it would not be a good time investment.

e1himself: Hi @kamilmielnik,

Thank you for your comment.

I really want to reopen this discussion.

I was fine with the frontend team deciding on conventions as long as it's only affecting the stuff *you* are working with. But [as I mentioned before](https://github.com/prezly/prezly/pull/9088/files#r551201217) it's the backend team working mainly on Prezly models interfaces, so I want this to be convenient for us to do our job. Having the props sorted alphabetically just makes it *much more difficult* for me to trace what has to be added or removed as we change the APIs. I really mean it: it is much more difficult.

I find it unfair that my opinion is discarded, while I spend much more time maintaining these files than you do.

src/types/index.ts
e1himself: Same reason *namespaces* or *modules* exist -- you group similar entities and separate those that differ. 

Contacts, Categories, Coverage, Newsroom and Stories -- they all are entities -- [first-class citizens](https://en.wikipedia.org/wiki/First-class_citizen) in our domain model -- they have their own API endpoints, they can be retrieved, updated and deleted by itself.

While *Pagination*, *Selection*, *OEmbedInfo* and *Uploadcare* files are not *entites*, but *value objects* --  *second-class objects* or just "secondary". They don't have they own APIs, they cannot be deleted or updated, they only appear as part of other entities APIs.

kamilmielnik: Ok, I get it - we have **entities** and **other**. 

I think that whatever word is used for **other**, it should (as much as possible):
1. explain the purpose
2. indicate that **entities** do not belong there
3. be an antonym of "entities" (well, it should be enough to just have a different meaning)

With "secondary" I ask myself: "secondary what?". And since there are **entities**, it makes me think they are "secondary entities", which is not the case.
I guess it's because I didn't figure out the "first-class citizen" analogy (maybe because I have never heard "second-class citizen").

Anyway, my failure to understand this leads me to these suggestions which I believe are a bit better than `types/secondary`:
- `types/common` (my pick)
- `types/utils` 
- `types/helpers`

src/Sdk/NewsroomContacts/types.ts
kamilmielnik: Great!

Let's have a call about this and sort it out once and for all.
How about next week, after we wrap up the website? (Wednesday?)

kamilmielnik (approved):

src/Sdk/NewsroomContacts/types.ts
e1himself: Scheduled a call for next Wednesday at 11:00 AM (BRU).

src/types/index.ts
e1himself: As of me, I think `secondary` is more specific, as *common* is pretty much overused everywhere. Also the top-level `types` folder is *common* by its own already. 

But I have renamed it to `common` just to not overspend here. Having some sub-folder is better than not having it. So it's an improvement already.

Thanks!

PR #23, opened 2021-01-26
[DEV-1571] Feature - Newsroom Subscriptions SDK

M4rk07:
- Support Newsroom Subscriptions API in SDK.

### DIFF
--- a/src/Sdk/NewsroomSubscriptions/NewsroomSubscriptions.ts
+++ b/src/Sdk/NewsroomSubscriptions/NewsroomSubscriptions.ts
@@ -1,7 +1,7 @@
-import ApiClient from "../ApiClient";
-import {Newsroom} from "../../types";
-import {NewsroomSubscriptionCreateRequest} from "./types";
-import routing from "../routing";
+import ApiClient from '../ApiClient';
+import { Newsroom } from '../../types';
+import { NewsroomSubscriptionCreateRequest } from './types';
+import routing from '../routing';
 
 export default class NewsroomSubscriptions {
     private readonly apiClient: ApiClient;
--- a/src/Sdk/NewsroomSubscriptions/NewsroomSubscriptions.ts
+++ b/src/Sdk/NewsroomSubscriptions/NewsroomSubscriptions.ts
@@ -1,0 +1,22 @@
+import ApiClient from "../ApiClient";
+import {Newsroom} from "../../types";
+import {NewsroomSubscriptionCreateRequest} from "./types";
+import routing from "../routing";
+
+export default class NewsroomSubscriptions {
+    private readonly apiClient: ApiClient;
+
+    constructor({ apiClient }: { apiClient: ApiClient }) {
+        this.apiClient = apiClient;
+    }
+
+    public async create(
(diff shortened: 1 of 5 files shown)


src/Sdk/NewsroomSubscriptions/types.ts
```diff
@@ -0,0 +1,38 @@
+import { Contact, UploadcareImage } from '../../types';
+
+export interface NewsroomSubscriptionCreateRequest {
+    email_address: string;
+    locale?: string;
+    url?: string | null;
```
kamilmielnik, line 6: Do `null` and `undefined` produce the same behavior? 
If yes, then I'd avoid using nulls.
If not, then it'd be nice to leave a comment explaining the difference.
This comment applies to all other similar cases.

kamilmielnik (approved):

src/Sdk/NewsroomSubscriptions/types.ts
```diff
@@ -0,0 +1,38 @@
...
+    ip_address?: string | null;
+    visitor_uid?: string | null;
+    session_uid?: string | null;
+    contact?: {
+        contact_type: Contact['contact_type'];
+        company_name: Contact['company_name'];
```
e1himself, line 12: `company_name` is only required when `contact_type === 'organisation'`

```suggestion
        company_name?: Contact['company_name'];
```

src/Sdk/NewsroomSubscriptions/types.ts
```diff
@@ -0,0 +1,38 @@
...
+        first_name?: Contact['first_name'];
+        last_name?: Contact['last_name'];
+        gender?: Contact['gender'];
+        function_name?: Contact['function_name'];
+        organisations?: string[];
```
e1himself, line 17: `first_name`, `last_name`, `gender`, `function_name` and `organisations` is only available when `contact_type === "person". I'd mention this in properties comments.

src/Sdk/NewsroomSubscriptions/NewsroomSubscriptions.ts
```diff
@@ -0,0 +1,22 @@
...
+import routing from '../routing';
+
+export default class NewsroomSubscriptions {
+    private readonly apiClient: ApiClient;
+
+    constructor({ apiClient }: { apiClient: ApiClient }) {
+        this.apiClient = apiClient;
+    }
+
+    public async create(
```
e1himself, line 13: I think it's worth to call this method `subscribe()`.

e1himself (commented):
:+1:

src/Sdk/NewsroomSubscriptions/types.ts
M4rk07: Yes, parameter not sent is the same as it was `null`, if that is what you asked? I removed the nulls. 👍

src/Sdk/NewsroomSubscriptions/types.ts
M4rk07: Nice suggestion, I added comments. Thanks!

PR #24, opened 2021-01-29
Feature - Type-hint Story `content` property

e1himself:
- Use `@prezly/slate-types` package to type hint v3 stories `content` prop

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1535,6 +1535,22 @@
                 "fastq": "^1.6.0"
             }
         },
+        "@prezly/sdk": {
+            "version": "2.4.0",
+            "resolved": "https://registry.npmjs.org/@prezly/sdk/-/sdk-2.4.0.tgz",
+            "integrity": "sha512-ZilGaJahuk3dGWtEOHpjvSaa05ykmpc8Ua6kOOv5EKRn7z1YqHX40z9BizdoOhS3AS9FWawuonKFPFsxIzXZRQ==",
+            "requires": {
+                "query-string": "^6.9.0"
+            }
+        },
+        "@prezly/slate-types": {
+            "version": "0.1.0-alpha.11",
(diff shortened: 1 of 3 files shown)


kamilmielnik (approved):

kamilmielnik:
"Check / Build, Test & Prettier (10.x)" - that GitHub action does not like seem to like SRP :)

If all we get errors from test phase & prettier phase, will we see both errors at once, or do we have to fix the first one, run build again, and only then be able to see the second error?

e1himself:
@kamilmielnik 

> "Check / Build, Test & Prettier (10.x)" - that GitHub action does not seem to like SRP :)

Indeed :)

> If we get errors from test phase & prettier phase, will we see both errors at once, or do we have to fix the first one, run build again, and only then be able to see the second error?

It stops on the first failed step:

https://github.com/prezly/javascript-sdk/blob/10e03e6b964a6981fdd8686031ab71bceab7baea/.github/workflows/nodejs.yml#L26-L36

This, however, can be changed to run them in parallel (see https://github.com/prezly/prezly/pull/9090 for example)

More importantly, I think you should look at this error:

> Error: node_modules/@prezly/slate-types/build/sdk/UploadcareFile.d.ts(1,26): error TS2307: Cannot find module 'uploadcare-widget' or its corresponding type declarations.

I guess this means that there is something wrong with the `@prezly/slate-types` package build.

kamilmielnik:
Thanks, I'll have a look at it today!

kamilmielnik:
I know what the problem is, but I still have to figure out the solution. I will continue tomorrow.

kamilmielnik:
@e1himself fixed with https://github.com/prezly/javascript-sdk/pull/24/commits/c97a91d583d174dd6fae995f9301fe68fd19f7dc :construction_worker:

Thank you for the report!

e1himself:
Thanks a lot for helping here @kamilmielnik :+1: :tada:

PR #25, opened 2021-03-30
[DEV-2388] Feature - Promote UUID

e1himself:
- Promote `uuid` as primary identifiers for:
   * Newsroom
   * NewsroomContact
   * Story
   * Coverage

### DIFF
--- a/src/Sdk/Coverage/Coverage.ts
+++ b/src/Sdk/Coverage/Coverage.ts
@@ -57,7 +57,10 @@ export default class CoverageSdk {
         return response.payload.coverage;
     }
 
-    async update(id: Coverage['uuid'] | Coverage['id'], payload: CoverageUpdateRequest): Promise<Coverage> {
+    async update(
+        id: Coverage['uuid'] | Coverage['id'],
+        payload: CoverageUpdateRequest,
+    ): Promise<Coverage> {
         const response = await this.apiClient.patch<{ coverage: Coverage }>(
             `${routing.coverageUrl}/${id}`,
             { payload },
--- a/src/Sdk/Coverage/Coverage.ts
+++ b/src/Sdk/Coverage/Coverage.ts
@@ -10,8 +10,6 @@ import {
     CoverageUpdateRequest,
 } from './types';
 
-type CoverageId = Coverage['id'];
-
 export default class CoverageSdk {
     private readonly apiClient: ApiClient;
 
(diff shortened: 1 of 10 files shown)


src/Sdk/Coverage/Coverage.ts
```diff
@@ -33,7 +31,7 @@ export default class CoverageSdk {
         return response.payload;
     }
 
-    async get(id: CoverageId, includeDeleted = false): Promise<Coverage> {
+    async get(id: Coverage['uuid'] | Coverage['id'], includeDeleted = false): Promise<Coverage> {
```
oleg-semyonov, line 34: Why not leaving the `CoverageId` type and setting it to `Coverage['uuid'] | Coverage['id']`, to make things more streamlined?

oleg-semyonov: And I guess same could be done for other types. This way, if we then decide to completely get rid of referencing by `id`, we would only need to update a single type.

src/types/Newsroom.ts
```diff
@@ -1,6 +1,10 @@
 import Culture from './Culture';
 
 export interface NewsroomRef {
+    /**
+     * @deprecated Please use `uuid` instead.
+     * @see uuid
+     */
     id: number;
     /**
      * UUID is going to replace numeric `id` in future.
```
oleg-semyonov, line 10: This comment looks like a duplicate of the newly created comment above.

oleg-semyonov (changes requested):

src/Sdk/Coverage/Coverage.ts
e1himself: My thinking was that `CoverageId` naming is confusing, as it refers to UUID or ID; with UUID being preferred. So I have decided to just inline the union type to make it maximum obvious which values it accepts.

Do you still think I should reintroduce `CoverageId` type here?

oleg-semyonov: I think it will make the maintaining of the types easier for sure.
Agree with the naming part, maybe call that `CoverageReference`? That's generic enough and still communicates the purpose of the type. (Though the argument is still called `id`, so maybe `CoverageId` is still OK 😁)

e1himself: Alright, I've re-introduced the `*Id` types. I think it's more straighforward than `CoverageReference`.

Thanks for your comments Oleg :+1:

ba8086e0b17793a4d6b4f09b60e721bb2fdc512b

oleg-semyonov (approved):

pablopunk (approved):

PR #32, opened 2021-04-15
[DEV-2559] Feature - Newsroom Webhooks API

M4rk07:
DEV-2559

Documentation of the API [here](https://developers.prezly.com/docs/api/reference/api/API.v2.yaml/paths/~1newsrooms~1%7Bnewsroom_id%7D~1webhooks/get).


src/types/NewsroomWebhook.ts
```diff
@@ -0,0 +1,8 @@
+export default interface NewsroomWebhook {
+    id: string;
+    name: string;
+    url: string;
+    events: string[];
```
e1himself, line 5: Let's introduce an enum for event names?

e1himself (approved):

src/types/NewsroomWebhook.ts
M4rk07: Thanks for the suggestion, done!

yuriyyakym (approved):
Looks good to me 👍

PR #43, opened 2021-06-07
[DEV-3261] Feature - Add NewsroomContacts get method

yuriyyakym:


src/Sdk/NewsroomContacts/NewsroomContacts.ts
```diff
@@ -11,6 +11,7 @@ import {
 } from './types';
 
 type NewsroomId = Newsroom['uuid'] | Newsroom['id'];
+type ContactId = NewsroomContact['uuid'] | NewsroomContact['id'];
```
kamilmielnik, line 14:
```suggestion
type NewsroomContactId = NewsroomContact['uuid'] | NewsroomContact['id'];
```

`NewsroomContact` is significantly different from a `Contact`, let's not introduce any confusion.

kamilmielnik (approved):

src/Sdk/NewsroomContacts/NewsroomContacts.ts
yuriyyakym: 37fbbc3

PR #46, opened 2021-06-07
Fix - UserRef interface

e1himself:
- Add missing props to `UserRef` interface
- Drop `UserAccountRef` interface
  Which was really just a desynchronized copy of `UserRef`. It is the same entity.
- Export `UserRef` interface

Related to #45

### DIFF
--- a/src/types/UserRef.ts
+++ b/src/types/UserRef.ts
@@ -1,16 +1,19 @@
 export default interface UserRef {
     id: number;
+    display_name: string;
+    avatar_url: string;
     /**
      * @deprecated Please use `email` instead.
      * @see email
      */
     username: string;
     email: string;
-    display_name: string;
     first_name: string | null;
-    avatar_url: string;
-    /**
-     * Last time the user was active. ISO-8601 format.
-     */
+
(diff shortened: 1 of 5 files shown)


src/types/UserRef.ts
```diff
@@ -1,16 +1,19 @@
...
-    display_name: string;
     first_name: string | null;
-    avatar_url: string;
-    /**
-     * Last time the user was active. ISO-8601 format.
-     */
+
+    is_terms_of_service_accepted: boolean;
+    sign_in_flow: 'google' | 'password' | 'sso';
+    use_case_answer: string;
```
kamilmielnik, line 15: It's an enum in `prezly`: https://github.com/prezly/prezly/blob/master/apps/backend/resources/javascripts/types/api/UseCaseAnswer.ts

kamilmielnik (approved):

src/types/UserRef.ts
e1himself: You're right. But should we expose a list of possible values to (3rd party) Prezly API consumers?

PR #63, opened 2021-08-09
Add "is_custom" field to NewsroomTheme interface

yuriyyakym:


src/types/NewsroomThemePreset.ts
```diff
@@ -15,10 +15,11 @@ interface EditableSettings {
 }
 
 export interface NewsroomTheme {
-    id: string;
-    name: string;
     description: string;
     features: string[];
+    id: string;
+    is_custom: boolean;
+    name: string;
```
e1himself, line 22: FWIW I do remember we had an agreement to not sort shape properties alphabetically.

:man_shrugging:

PR #68, opened 2021-09-01
Feature - Support UTM parameters in Unsubscribe APIs

marko-ogg:


src/Sdk/Subscriptions/types.ts
```diff
@@ -37,7 +38,11 @@ export interface SubscribeRequest<Type extends ContactType> {
 }
 
 export interface UnsubscribeRequest {
-    email_address: string;
+    /**
+     * Either utm.utm_id or email_address must be provided
+     */
```
oleg-semyonov, line 43: @marko-ogg Does that mean that we can have an unsubscribe link only with `utm_id` and no `email` parameter?

oleg-semyonov (approved):

src/Sdk/Subscriptions/types.ts
marko-ogg: Yes. In that case, we will try to search for campaign recipient by that `utm_id`, and if we can find him, he will be unsubscribed.

oleg-semyonov: Okay, thanks!

marko-ogg: @oleg-semyonov Also, it's not a problem if you provide `utm_id` and `email` at the same time. We will try to find recipient by that `utm_id`, and if it matches the provided `email`, we will still consider it as "unsubscribe from campaign". If provided email doesn't match recipient's email that is provided with `utm_id`, then we will just process that request as a normal "email unsubscribe" request.

PR #69, opened 2021-09-13
[DEV-3425] Feature - Share newsroom domain dns instructions

yuriyyakym:


yuriyyakym:
In email domains instructions, urls were included in domains response. For newsrooms domains we need to make additional request. I don't know why do we have different approaches. But I like current approach more, since token is generated on-demand only.

e1himself:
> In email domains, instructions urls were included in domains response. For newsrooms domains we need to make additional request. I don't know why do we have different approaches. But I like current approach more, since token is generated on-demand only.

That's something I didn't like in the initial implementation of Sender Domains API either. But we've decided to take that approach for the sake of simplicity. "Some day" :tm: we'll fix it.

:man_shrugging:

e1himself (approved):

PR #82, opened 2021-11-01
Feature - Add slug to StoryRef

marko-ogg:


oleg-semyonov (approved):
So the field was already there, only the typing were missing? 🤔

marko-ogg:
@oleg-semyonov no, I also updated the Prezly API.

PR #87, opened 2021-11-05
Refactor - Avoid default exports

e1himself:
- Use named exports in most of the modules
- Leverage wildcard export (`export *`) to reduce the amount of repetitive code
  - ~300 LOC less
  - as another advantage we will not be forgetting to re-export new types, as most of stuff is exported by default
- Move secondary types around closer to their consuming code
- Rename `WebhookEvent` to `NewsroomWebhookEvent` (BC break)


yuriyyakym (approved):
❤️

kudlajz (approved):

PR #88, opened 2021-11-08
Feature - Support Newsroom Galleries API

marko-ogg:


kudlajz (changes requested):
Looks good 👍  but I believe you should not use default exports anymore, see https://github.com/prezly/javascript-sdk/pull/87

kudlajz (approved):
👏

PR #92, opened 2021-11-09
Stringify scope and query when doing galleries list request

marko-ogg:


kudlajz (approved):

e1himself:
@marko-ogg FYI I've fixed the `query` type requirements to avoid this class of mistakes in future: e6c4843f4.

PR #96, opened 2021-11-19
Fix - Missing Utm interface properties

marko-ogg:
Oversight from this PR (after moving `Utm.ts` to `common`): https://github.com/prezly/javascript-sdk/pull/87


src/types/common/index.ts
```diff
@@ -2,4 +2,5 @@ export * from './Notification';
 export * from './OEmbedInfo';
 export * from './Pagination';
 export * from './Query';
+export * from './Utm';
```
e1himself, line 5: Actually in #87 I've moved it to Subscriptions SDK, as it was a local type anyway (not publicly exported from `types` folder).

https://github.com/prezly/javascript-sdk/blob/269a98f3a1ca10e0b621f0e99f9767401689b0ee/src/Sdk/Subscriptions/types.ts#L3-L9

So please remove the dupe definition of it in `Subscriptions/types.ts`.

e1himself (approved):

src/types/common/index.ts
marko-ogg: But you also left empty `Utm.ts` file in `common`: https://github.com/prezly/javascript-sdk/blob/269a98f3a1ca10e0b621f0e99f9767401689b0ee/src/types/common/Utm.ts

e1himself: :open_mouth:

PR #107, opened 2022-01-18
Feature - Implement newsroom privacy requests SDK

marko-ogg:


src/Sdk/NewsroomPrivacyRequests/NewsroomPrivacyRequests.ts
```diff
@@ -0,0 +1,40 @@
+import { Newsroom, PrivacyRequest } from '../../types';
+
+import DeferredJobsApiClient from '../DeferredJobsApiClient';
+import routing from '../routing';
+import { NewsroomPrivacyRequestCreateRequest } from './types';
+
+type NewsroomId = Newsroom['uuid'] | Newsroom['id'];
```
oleg-semyonov, line 7: Since this is a totally new SDK method, shouldn't we force the usage of `uuid` here?
Or are there some scenarios, where only `id` can be provided?

src/Sdk/NewsroomPrivacyRequests/types.ts
```diff
@@ -0,0 +1,15 @@
...
+    type: PrivacyRequestType;
+    /**
+     * Required for "delete", "correct" and "other"
+     * privacy request types
+     */
+    message?: string;
+    /**
+     * Required for "correct" privacy request type
+     */
+    extra_message?: string;
```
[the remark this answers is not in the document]
marko-ogg: You are right. I've created separate privacy request create request types. Please check out if that is what you meant.

src/Sdk/NewsroomPrivacyRequests/NewsroomPrivacyRequests.ts
marko-ogg: No, I just wrote it based on how we use it in other SDKs. But makes sense, we should be getting rid of deprecated `id`. Thanks! 👍🏻

src/Sdk/NewsroomPrivacyRequests/types.ts
```diff
@@ -0,0 +1,29 @@
...
+    message: string;
+    extra_message: string;
+}
+
+export interface OtherPrivacyRequestCreateRequest {
+    email: string;
+    type: PrivacyRequestType;
+    message: string;
+    extra_message?: string;
+}
```
kudlajz, line 29: Should it be like this?

```suggestion
export interface DeletePrivacyRequestCreateRequest {
    email: string;
    type: PrivacyRequestType.DELETE;
    message: string;
    extra_message?: string;
}

export interface ExportPrivacyRequestCreateRequest {
    email: string;
    type: PrivacyRequestType.EXPORT;
    message?: string;
    extra_message?: string;
}

export interface CorrectPrivacyRequestCreateRequest {
    email: string;
    type: PrivacyRequestType.CORRECT;
    message: string;
    extra_message: string;
}

export interface OtherPrivacyRequestCreateRequest {
    email: string;
    type: PrivacyRequestType.OTHER;
    message: string;
    extra_message?: string;
}
```

src/Sdk/NewsroomPrivacyRequests/NewsroomPrivacyRequests.ts
```diff
@@ -0,0 +1,47 @@
...
+        const { privacy_request } = await this.apiClient.post<{ privacy_request: PrivacyRequest }>(
+            url,
+            {
+                payload,
+            },
+        );
+        return privacy_request;
+    }
+
+    public async confirm(
```
kudlajz, line 37: We're not gonna actually use this method, right?

Currently the data request is confirmed by clicking a link in the email as far as I know and then you're redirected to the newsroom's "confirmation successful" page.

Does this approach stay the same? We don't have such screen in Figma, should we add it?

kudlajz (approved):
Looks good 👍  I've left some questions though as I'm not sure about the confirmation.

src/Sdk/NewsroomPrivacyRequests/NewsroomPrivacyRequests.ts
marko-ogg: I think this should be handled similar as for subscribing. We are sending email from privacy portal, not from backend.

So the steps (from privacy portal) would be:
1. Create privacy request
2. Send confirmation email
3. When confirmation link is clicked, send `confirm` request to API and display success message

Does this make sense or you have a different suggestion?

src/Sdk/NewsroomPrivacyRequests/types.ts
marko-ogg: Totally! Thanks 👌🏻

src/Sdk/NewsroomPrivacyRequests/NewsroomPrivacyRequests.ts
kudlajz: Yeah, it makes sense. So I'll have to add the confirmation page as well then. This can be merged 👍

PR #115, opened 2022-04-07
[TF-661] Expose the Segment Write Key in the Newsroom API object

pestaa:
Together with:

* https://github.com/prezly/prezly/pull/10874/files

### DIFF
--- a/src/types/Newsroom.ts
+++ b/src/types/Newsroom.ts
@@ -93,6 +93,7 @@ export interface Newsroom extends NewsroomRef {
         category: string | null;
     };
     ga_tracking_id: string | null;
+    segment_analytics_id: string | null;
 
     is_subscription_form_enabled: boolean;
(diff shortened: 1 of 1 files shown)


src/types/Newsroom.ts
```diff
@@ -93,6 +93,7 @@ export interface Newsroom extends NewsroomRef {
         category: string | null;
     };
     ga_tracking_id: string | null;
+    segment_analytics_id: string | null;
```
oleg-semyonov, line 96: So this property was there all along? 🤔

oleg-semyonov (approved):

src/types/Newsroom.ts
pestaa: It wasn't exposed, like you mentioned. The column and PATCH API were implemented.

PR #117, opened 2022-04-26
Improve CI performance and fix TS errors

aspirisen:
Discussion https://prezly.slack.com/archives/C029LTE0M/p1651013990025259


src/Api/createRequest.ts
```diff
@@ -111,9 +111,9 @@ export default async function createRequest<P = any>(
         // Since we also throw when response is not ok, re-throw the response data if available.
         const {
             status,
-            statusText = error.message || NETWORK_PROBLEM_ERROR_MESSAGE,
+            statusText = (error as any).message || NETWORK_PROBLEM_ERROR_MESSAGE,
```
aspirisen, line 114: Looks like this package is built with old TS version where `error` in catch has `any` type

e1himself (approved):
:+1:

PR #124, opened 2022-06-06
Support - Upgrade node requirements

e1himself:
- [Run CI tests on node v16 only](https://github.com/prezly/javascript-sdk/commit/aebeeae18df56060381bf694a98c1dd2c8f1b4bd)
- [Upgrade package-lock.json to the newer v2 format](https://github.com/prezly/javascript-sdk/commit/47685b63ae6d46836937fdafe9f27c3c563794b7)

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -10,7 +10,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x]
+        node-version: [16.x]
 
     env:
       CI: true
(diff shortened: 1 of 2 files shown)


oleg-semyonov (commented):
I see that `package.json` still [mentiones](https://github.com/prezly/javascript-sdk/blob/9c1d32c9856a1efbea1da72ebe67c8a618684fe6/package.json#L15-L17) a really old version in its "engines" property:
```
"engines": {
    "node": ">= 8.16.0"
},
```

Also worth specifying a preferred node version in an `.nvmrc` file 🤔

PR #145, opened 2022-10-11
Define additional gallery API entity properties

marko-ogg:
- name
- description
- thumbnail_image
- creator
- modified_at
- url

### DIFF
--- a/src/types/NewsroomGallery.ts
+++ b/src/types/NewsroomGallery.ts
@@ -1,7 +1,7 @@
 import { GalleryStatus } from './GalleryStatus';
 import { NewsroomGalleryImage } from './NewsroomGalleryImage';
-import {UserRef} from "./UserRef";
-import {UploadedImage} from "@prezly/uploads";
+import { UserRef } from './UserRef';
+import { UploadedImage } from '@prezly/uploads';
 
 export interface NewsroomGallery {
     id: number;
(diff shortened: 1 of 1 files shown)


src/types/NewsroomGallery.ts
```diff
@@ -1,13 +1,22 @@
...
+import { UploadedImage } from '@prezly/uploads';
 
 export interface NewsroomGallery {
     id: number;
     uuid: string;
+    name: string;
+    /** @deprecated Use `name` instead */
     title: string;
+    description: string | null;
+    thumbnail_image: UploadedImage | null;
```
kudlajz, line 13: I've just spent 30 minutes debugging why I can't create `UploadcareImage` from this property and it turns out that the API sends a stringified JSON instead of actual object as typed here 😬 Can you please fix that in the API?

marko-ogg: Ah, I see, sorry about that! But I did the API part same as it works with stories, you get the json string for `thumbnail_image`. Do you think that maybe just changing the spec here make sense to keep it consistent?

kudlajz: I don't really care which one is it so it's up to you 🙂

marko-ogg: Released under `7.11.0-1`.

PR #151, opened 2022-10-19
[DEV-9064] Refactor - Exported symbols naming

e1himself:
- Refactor `Sdk` client
   - Rename to `Client` 
   - Change `Sdk` from default export to named `PrezlyClient` type
   - Change from class to a POJO object + factory function
- Reorganize folder structure
- Refactor how exports are grouped per endpoints: it's one ES namespace for every endpoint
  - Endpoint client class is always called `Client` now (e.g. `Newsrooms.Client`)
  - Endpoint-specific types are not prefixed with endpoint name, but rely on ES namespace instead (`NewsroomsListResponse` &rarr; `Newsrooms.ListResponse`)
- Unify naming and behavior for List / Search APIs endpoints 
   - Unify their types naming: `ListOptions` / `SearchOptions` (previously: `ListRequest`, `ListOptions`, `ListRequestOptions`, `SearchRequest`, `SearchOptions`, `SearchRequestOptions`)
   - List APIs never contain query object, it's only available for Search APIs
   - Search query object is always named as `query` now (previously: `jsonQuery` and `query`)
   - Use `limit` / `offset` parameters for all List / Search APIs (previously: `page`, `pageSize`, `limit`, and `offset`)
   - Make all `query` params accept both string and object


src/endpoints/Coverage/types.ts
```diff
@@ -4,16 +4,17 @@ import {
     CultureRef,
     CountryRef,
     Contact,
-    Coverage,
+    CoverageEntry,
```
e1himself, line 7: Had to rename this to fix naming collision: `Coverage` endpoint ES namespace vs `Coverage` entity

aspirisen (approved):

oleg-semyonov (approved):
Wow, this is huge! Great work!



Since we're introducing such big changes, I wonder if it would make sense to try adding ES Module version of the package prior to releasing the SDK v8 (not in scope of this PR of course)? 
I know that some of the dependencies are quite old and they in turn might need updating to ES Modules, but this might be a great opportunity to bring our packages to modern standards. Do you think it's feasible or it's better to do in v9 or later? 🤔

PR #156, opened 2022-10-19
Dev - ESM build

e1himself:
Switch package to ESM-only build


aspirisen (approved):
👍

oleg-semyonov:
Shouldn't this have been merged into `next` branch though? 🤔 
We might encounter some issues with consumer apps that run on CommonJS, so this change should land in a major update

e1himself:
> We might encounter some issues with consumer apps that run on CommonJS, so this change should land in a major update

Released with a major version update: v9.0.0

PR #160, opened 2022-10-25
[QRF-350] Fix typo in Periodicity.BI_ANUALLY enum value

oleg-semyonov:


src/types/Contact.ts
```diff
@@ -24,7 +24,7 @@ export enum Periodicity {
     BI_MONTHLY = 'bi-monthly',
     QUARTERLY = 'quarterly',
     ANNUALLY = 'annually',
-    BI_ANUALLY = 'bi-anually',
+    BI_ANNUALLY = 'bi-annually',
```
e1himself, line 27: > `BI_ANNUALLY`

The constant name cannot be renamed under 7.x. As it's a BC break. Let's only change the value here.

oleg-semyonov: Oh, right, my bad

oleg-semyonov: https://github.com/prezly/javascript-sdk/pull/160/commits/039ce32f80a029ffd39b1dec0ec43eca57341c31

e1himself (approved):

e1himself:
This patch has been propagated to 8.x and 9.x as: v8.0.1 and v9.0.2

PR #161, opened 2022-10-26
Feature - Migrate more types from the app

e1himself:
Needed for https://github.com/prezly/prezly/pull/11668

### DIFF
--- a/src/endpoints/CampaignRecipients/scopes.ts
+++ b/src/endpoints/CampaignRecipients/scopes.ts
@@ -1,30 +1,30 @@
 import type { Campaign, Contact, EmailRecipient, SelectionValue } from '../../types';
 
 export interface AllContactsScope {
-    type: 'scope:contacts';
+    type: ContactsScope.Type.ALL_CONTACTS;
     selection?: SelectionValue<Contact['id']>;
 }
 
 export interface ContactOrganisationsScope {
-    type: 'scope:contact_organisations';
+    type: ContactsScope.Type.CONTACT_ORGANISATIONS;
     contact_id: Contact['id'];
     selection?: SelectionValue<Contact['id']>;
 }
 
 export interface ContactEmployeesScope {
-    type: 'scope:contact_organisations';
+    type: ContactsScope.Type.CONTACT_EMPLOYEES;
     contact_id: Contact['id'];
     selection?: SelectionValue<Contact['id']>;
 }
 
 export interface CampaignRecipientsScope {
-    type: 'scope:campaign_recipients';
+    type: ContactsScope.Type.CAMPAIGN_RECIPIENTS;
     campaign_id: Campaign['id'];
     selection?: SelectionValue<EmailRecipient['id']>;
 }
 
 export interface CampaignReportScope {
-    type: 'scope:campaign_recipients';
+    type: ContactsScope.Type.CAMPAIGN_REPORT;
     campaign_id: Campaign['id'];
     report:
         | 'recipients'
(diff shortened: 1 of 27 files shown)


src/endpoints/CampaignRecipients/scopes.ts
```diff
@@ -1,30 +1,30 @@
...
 
 export interface CampaignRecipientsScope {
-    type: 'scope:campaign_recipients';
+    type: ContactsScope.Type.CAMPAIGN_RECIPIENTS;
     campaign_id: Campaign['id'];
     selection?: SelectionValue<EmailRecipient['id']>;
 }
 
 export interface CampaignReportScope {
-    type: 'scope:campaign_recipients';
```
e1himself, line 27: Fun fact: this was using a wrong `type`

src/endpoints/CampaignRecipients/scopes.ts
```diff
@@ -1,30 +1,30 @@
...
 
 export interface ContactOrganisationsScope {
-    type: 'scope:contact_organisations';
+    type: ContactsScope.Type.CONTACT_ORGANISATIONS;
     contact_id: Contact['id'];
     selection?: SelectionValue<Contact['id']>;
 }
 
 export interface ContactEmployeesScope {
-    type: 'scope:contact_organisations';
```
e1himself, line 15: Fun fact: this was using a wrong `type`

src/types/Contact.ts
```diff
@@ -3,8 +3,9 @@ import type { UploadedImage } from '@prezly/uploads';
 import type { ContactDuplicateSuggestion } from './ContactDuplicateSuggestion';
 import type { Entity } from './Entity';
 
-export interface ContactRef extends Entity {
-    contact_type: 'person' | 'organisation';
+export interface ContactRef extends Entity<number> {
+    id: number;
```
oleg-semyonov, line 7: Is this needed, if `id` property is inherited from the `Entity` type? Same on line 20

OskarGasiorowski (approved):

src/types/License.ts
```diff
@@ -9,16 +11,99 @@ export interface LicenseRef {
...
+    canceled_at: string | null;
+    company_address: string;
+    company_address_city: string;
+    company_address_country: string;
+    company_address_zip: string;
     created_at: string;
+    /**
+     * When there are no subscriptions yet, the currency is null (we do not set a default value).
+     */
+    currency: 'eur' | 'usd' | 'gbp' | null;
```
oleg-semyonov, line 39: This looks like something that could be extracted into an enum :)

src/types/Entity.ts
```diff
@@ -1,3 +1,3 @@
-export interface Entity<T extends number | string = number> {
+export interface Entity<T extends number | string = string | number> {
```
oleg-semyonov, line 1: I don't see any `Entity` extends without specified type in this PR, which makes me a bit confused about this change.

Do we even have `string` IDs anywhere? 🤔 
I was under impression that all `id` fields are numbers, with strings left for `uuid`.

e1himself: @oleg-semyonov Entity<string> is relied on in the main app. Also, `EmailRecipient` is de facto extending `Entity<string>`, though this is not reflected in the declaration (I'll add it, to be clear).

Thanks!

src/types/License.ts
e1himself: Will do. Thanks!

src/types/Contact.ts
e1himself: Well, it's not really necessary. But it also makes things more clear -- it's immediately obvious that there's `id` property. So I'm hesitant to remove it.

src/types/License.ts
e1himself: 434d8bd7faf939eb99a05999d972f4690a433cf9

src/types/Entity.ts
e1himself: 90212f8f53407bda0c92b022dff8994351e3d541

e1himself: Hm... I'm now thinking maybe we don't even need `Entity` type in this repository. It's not adding anything to *this* codebase.

MohammadxAli (approved):

src/types/Entity.ts
e1himself: > I'm now thinking maybe we don't even need Entity type in this repository. It's not adding anything to this codebase.

Dropped with 85323871ac9888a597cc9a9e67fa41597ff8ff34

src/types/Contact.ts
oleg-semyonov: I was about to say that in that case you might as well ditch the `Entity` interface, but I see you already did :)

PR #162, opened 2022-11-02
[DEV-9120][DEV-9121][DEV-9122][DEV-9123] Feature - Stories API

e1himself:
DEV-9120, DEV-9121, DEV-9122, DEV-9123

Migrate & improve Stories API functions from the application repo

### DIFF
--- a/src/endpoints/Stories/types.ts
+++ b/src/endpoints/Stories/types.ts
@@ -52,6 +52,8 @@ interface GenericCreateRequest {
 
     categories?: Category['id'][];
     tags?: string[];
+
+    is_shared_to_prpro?: boolean;
 }
 
 interface GenericUpdateRequest {
@@ -68,6 +70,8 @@ interface GenericUpdateRequest {
 
     categories?: Category['id'][];
     tags?: string[];
+
+    is_shared_to_prpro?: boolean;
 }
 
 export interface HtmlStoryCreateRequest extends GenericCreateRequest {
--- a/src/endpoints/Stories/types.ts
+++ b/src/endpoints/Stories/types.ts
@@ -115,10 +115,21 @@ export interface SlateStoryUpdateRequest extends GenericUpdateRequest {
      */
     intro?: never;
     content?: PrezlyContentFormat;
+    autosaved_content?: PrezlyContentFormat;
+    content_version?: number;
 }
 
 export type CreateRequest = HtmlStoryCreateRequest | SlateStoryCreateRequest;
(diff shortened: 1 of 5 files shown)


src/endpoints/Stories/types.ts
```diff
@@ -104,9 +113,76 @@ export interface SlateStoryUpdateRequest extends GenericUpdateRequest {
...
+    publish_at?: Iso8601DateTime;
+    visibility?: Story.Visibility.PUBLIC | Story.Visibility.EMBARGO;
+}
+
+export type UnscheduleRequest = UnpublishRequest;
+
+const EXTENDED_STORY_INCLUDED_ONDEMAND_FIELDS_SHAPE /*: Record<
+    keyof Omit<ExtendedStory, keyof Story>,
+    boolean
+>*/ = {
```
oleg-semyonov, line 153: Why couldn't you use the type here? Even if `satisfies` operator is not yet ready, you can still use the checks from implicit typing.

src/types/Story.ts
```diff
@@ -133,13 +177,7 @@ export namespace Story {
...
     }
 
-    export interface ExtraFields {
-        /**
-         * Depending on `format_version` this field can contain:
-         * - HTML content for v1 stories (deprecated)
-         * - JSON-encoded structured content for v3 stories (see Prezly Content Format).
-         */
-        content: string;
+    export interface OnDemandFields {
```
oleg-semyonov, line 180: I can already see all the fun we'll have with updating SDK in theme-kit and on the website-related repos, considering this will be another major release :D

oleg-semyonov (approved):

src/endpoints/Stories/types.ts
e1himself: :+1: Indeed. I've been trying this, but for some reason, it didn't work before. Tried again, and it's good. :man_shrugging: 

Will change. Thanks!

src/types/Story.ts
e1himself: But can you not just wait for the next important update and perform the necessary changes only when it's needed?

e1himself: My reasoning for this change is that `ExtendedStory` does not include all `ExtraFields`. Emphasizing the difference between `ExtendedStory` and on-demand includes felt like the right thing to do.

However, I can revert this change. If that will save a lot of time.

e1himself: 1a9a038e5f56cd91777c7022e1c538b0166ef9ea

MohammadxAli (approved):

OskarGasiorowski (approved):

PR #163, opened 2022-11-02
Feature - Implement notification subscriptions API SDK

marko-ogg:


src/types/NotificationType.ts
```diff
@@ -0,0 +1,6 @@
+export interface NotificationType {
+    id: string;
+    title: string;
+    description: string | null;
+    area: string;
+}
```
oleg-semyonov, line 6: I saw you using pre-defined constants in the backend code for `id` and `title`, but I assume these can be any value later on?

I would think that at least `area` should be constrained to a set of pre-defined values, since that won't change often, and will reduce the probability of errors.

oleg-semyonov (approved):

PR #166, opened 2022-11-07
[DEV-9210] Declare Change Newsroom API function

e1himself:


OskarGasiorowski (approved):
Good job @e1himself 🎉 
I think I can implement hook on app side in scope of this: https://linear.app/prezly/issue/DEV-9189/implement-change-newsroom-logic

PR #171, opened 2022-11-08
Fix `NotificationSubscriptions.UpdateRequest` type

oleg-semyonov:
This issue was introduced in #164. The API expects just the string id in the `notification_type` field, not the whole Type object

### DIFF
--- a/src/endpoints/NotificationSubscriptions/types.ts
+++ b/src/endpoints/NotificationSubscriptions/types.ts
@@ -2,7 +2,7 @@ import type { NotificationSubscription } from '../../types';
 
 interface SubscriptionUpdate {
     notification_type: NotificationSubscription.Type['id'];
-    is_email_subscribed: NotificationSubscription['is_email_required'];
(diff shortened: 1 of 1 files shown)


src/endpoints/NotificationSubscriptions/types.ts
```diff
@@ -1,5 +1,10 @@
 import type { NotificationSubscription } from '../../types';
 
+interface SubscriptionUpdate {
+    notification_type: NotificationSubscription.Type['id'];
+    is_email_subscribed: NotificationSubscription['is_email_required'];
```
marko-ogg, line 5: Here it should be `is_email_subscribed`.

e1himself (approved):

marko-ogg (approved):
👍🏻 see 1 remark below.

src/endpoints/NotificationSubscriptions/types.ts
oleg-semyonov: Oh snap! Autocomplete played some nasty tricks on me :D
Thanks for the catch!

PR #172, opened 2022-11-09
[DEV-9202] Fix inaccurate warning type for Story Change Newsroom API

e1himself:
https://github.com/prezly/prezly/pull/11775#discussion_r1017880044

### DIFF
--- a/src/endpoints/Stories/types.ts
+++ b/src/endpoints/Stories/types.ts
@@ -8,7 +8,6 @@ import type {
     Pagination,
     Query,
     Story,
-    Warning as BaseWarning,
 } from '../../types';
 
 /**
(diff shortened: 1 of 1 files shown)


OskarGasiorowski (approved):

src/endpoints/Stories/types.ts
```diff
@@ -25,6 +24,13 @@ type Html = string;
  */
 type PrezlyContentFormat = string;
 
+interface BaseWarning {
+    text: string;
+    scope: string;
+    field: string;
+    value: unknown;
+}
```
e1himself, line 32: > Just wondering about one thing. It's named BaseWarning but it's used only for changing newsrooms. Maybe it should be named more suggestive? Or we can use it in wider scope?

The Story API backend is using the same warning interface for other warning types too. Not just for Changing newsroom. 

That's why I think the name is good enough :)

OskarGasiorowski: In that case it make sense :)

PR #174, opened 2022-11-29
[DEV-9390] Add requests for Signup page

aspirisen:


src/endpoints/Signup/types.ts
```diff
@@ -0,0 +1,7 @@
+export type EmailAvailability = 'available' | 'taken';
```
oleg-semyonov, line 1: I think this should be an enum 🤔

src/endpoints/Billing/types.ts
```diff
@@ -0,0 +1,18 @@
+export interface SignupRequest {
+    email: string;
+    company_name: string;
+    first_name: string;
+    last_name: string;
+    plan: 'core';
+    currency: 'eur';
+    quantity: number;
+    billing_cycle: 'month';
```
oleg-semyonov, line 9: `plan`, `currency` and `billing_cycle` shouldn't be hardcoded. We were asked to provide ways to allow to sign up for any plan.

oleg-semyonov (changes requested):

src/endpoints/Billing/types.ts
oleg-semyonov: @aspirisen this one still stands. [Here is the code](https://github.com/prezly/prezly/blob/86312d46d76157f2b5d6a9714a13423522af1f83/lib/Prezly/Support/Plans/Plan.php#L28) that describes all supported values for `plan`, `currency` and `billing_cycle`

src/endpoints/Billing/types.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export enum SignupCurrencies {
+    EUR = 'eur',
+    USD = 'usd',
+}
+
+export enum SignupBillingCycles {
+    YEAR = 'year',
+    MONTH = 'month',
+}
```
marko-ogg, line 15: These are enum values, I wouldn't use plural, so: `SignupCurrency`, `SignupBillingCycle`. Maybe even without the `Signup` prefix.

marko-ogg (approved):

src/endpoints/Billing/types.ts
aspirisen: We already have such enums like here https://github.com/prezly/javascript-sdk/blob/master/src/types/Plan.ts
but looks like they do not much with this code https://github.com/prezly/prezly/blob/86312d46d76157f2b5d6a9714a13423522af1f83/lib/Prezly/Support/Plans/Plan.php#L28

can i use enums from shared types, or i have to create my own for signup?

aspirisen: hm, hence i have to create my own enums, thanks)

aspirisen:
@oleg-semyonov could you take a look

src/endpoints/Billing/types.ts
```diff
@@ -1,12 +1,28 @@
...
     company_name: string;
     first_name: string;
     last_name: string;
     quantity: number;
-    plan?: 'core';
-    currency?: 'eur';
-    billing_cycle?: 'month';
+    plan?: SignupPlan;
+    currency?: SignupCurrency;
+    billing_cycle?: SignupBillingCycle;
```
oleg-semyonov, line 25: Just to be sure - these are not required in the API, right? I saw them being required in the first commits 🤔

oleg-semyonov (approved):

src/endpoints/Billing/types.ts
aspirisen: well, technically they are mandatory
image

i thought that should be changed in future?

PR #177, opened 2022-12-06
Add newsroom pitches and coverages number properties

marko-ogg:


src/types/Newsroom.ts
```diff
@@ -79,6 +79,8 @@ export interface Newsroom extends NewsroomRef {
     cultures: CultureRef[];
     campaigns_number: number;
     stories_number: number;
+    pitches_number: number;
+    coverages_number: number;
```
kudlajz, line 83: I think you just want to make it consistent with other fields but coverage in this case is non-countable so I think `coverage_number` would be better, but I don't know 😅

kudlajz (approved):

src/types/Newsroom.ts
kudlajz: I guess it doesn't matter anymore since it's the API that does the naming 😄

marko-ogg: You were right, checked with Al and we should use `coverage_number` instead. 👍🏻 Refactored.

PR #178, opened 2022-12-06
[DEV-9459] Feature - Story Pinning API

e1himself:


fgyimah (approved):
👍🏾

src/endpoints/Stories/Client.ts
```diff
@@ -371,6 +371,38 @@ export class Client {
         return story;
     }
 
+    async pin<
+        Include extends readonly (keyof Story.ExtraFields)[],
+        Options extends IncludeOptions<Include>,
+        StoryRecord extends ExtendedStory = Options['include'] extends Include
+            ? ExtendedStory & Pick<Story.ExtraFields, Options['include'][number]>
+            : ExtendedStory,
```
MohammadxAli, line 379: This stuff is repeated in almost all requests but I know you probably have a good reason to keep it this way :)

But abstracting it and giving it a name would make it a little more readable maybe?

(Honestly, I don't know what this is, This is a world-class TypeScript for me.)

MohammadxAli (approved):

src/endpoints/Stories/Client.ts
e1himself: > Honestly, I don't know what this is

This is needed to allow TypeScript automatically infer the returned object shape from the passed options.

```ts
const story = client.stories.get(id); 
console.log(story.content); // TS error: object does not have `content` property

const extendedStory = client.stories.get(id, { include: ['content'] });
console.log(extendedStory.content); // OK
```

--- 

You're right that there's a lot of repetition here. But I don't really know how to abstract it at the moment, and don't really want to invest time into it now :) If you find a good way to avoid repeating it again and again -- I'd be happy to see it :)

e1himself: https://github.com/prezly/javascript-sdk/pull/263

PR #185, opened 2022-12-12
Feature - Support site SEO settings

marko-ogg:


src/types/NewsroomLanguageSettings.ts
```diff
@@ -14,4 +15,5 @@ export interface NewsroomLanguageSettings {
     default_cookie_statement: string;
     default_subscribe_disclaimer: string;
     notifications: Notification[];
+    seo_settings: SEOSettings;
```
marko-ogg, line 18: Note that this is a different API to get the language SEO settings (Newsroom Languages API), than it is used to update this settings (Newsrooms API).

Reason why I did this:
1. Single `Save` button on frontend side, so doesn't make sense to do two separate API requests to update the whole form (one to Newsrooms API second to Newsroom Languages API).
2. It would add to the newsroom payload if we would introduce a property that would contain settings per language.

@aspirisen let me know what you think of this and if this is ok to use on frontend?

aspirisen: Am i right that when we need to update settings seo page i will use Newsroom client and call `update` method with data from the the settings page using `google_search_console_key` and `cultures_seo_settings` from UpdateRequest type.

When i load the page i need to use  NewsroomLanguages client and call `get` method ?

Also, what field i should use to save and load this flag?
image

marko-ogg: @aspirisen that is correct, as you described. Would that be a problem that there is a different API to get this data? You think it's not a good practice, or have proposal to do it differently?

> Also, what field i should use to save and load this flag?

It's `is_indexable` and it already exists. We are using it in newsroom advanced settings, so we are basically just moving it as I understood.
Screenshot 2022-12-12 at 14 19 49

aspirisen: > Would that be a problem that there is a different API to get this data?
i think that's ok

one problem is that i will need to make two requests to load the page, one for `is_indexable` and another one to load everything else, but i think it is not a problem

aspirisen (approved):

src/endpoints/NewsroomLanguages/types.ts
```diff
@@ -12,7 +12,33 @@ export interface ListResponse {
 export interface SettingsUpdateRequest {
     is_default?: true;
     code?: CultureRef['code'];
-    company_information?: Partial<NewsroomCompanyInformation>;
+    company_information?: {
```
aspirisen, line 15: maybe we should create a separate type for these fields, but it is up to you

marko-ogg: I think it's fine for now since this is the only place where it is necessary. If the need arises for more places, we'll extract it into a separate type.

PR #186, opened 2022-12-13
[DEV-9503] Feature - Multi-column sorting

e1himself:
Followup on https://github.com/prezly/prezly/pull/12045

- Introduce `SortOrder` type and utilities
- Declare that APIs now support multi-column sorting


fgyimah (approved):
👍🏾

PR #188, opened 2022-12-26
[DEV-9713] Update - Newsroom Hub API

e1himself:
- Update Hub API methods to allow additional settings on hub members
- Also, it's now allowed to link/unlink specific members, instead of doing the whole list at once

### DIFF
--- a/src/endpoints/NewsroomHub/Client.ts
+++ b/src/endpoints/NewsroomHub/Client.ts
@@ -2,7 +2,7 @@ import type { DeferredJobsApiClient } from '../../api';
 import { routing } from '../../routing';
 import type { NewsroomRef } from '../../types';
 
-import type { ListResponse, LinkRequest, LinkResponse } from './types';
+import type { ListResponse, LinkOptions, HubMember, LinkManyRequest } from './types';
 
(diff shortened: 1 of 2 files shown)


src/endpoints/NewsroomHub/Client.ts
```diff
@@ -13,15 +13,48 @@ export class Client {
...
         const { members } = await this.apiClient.get<ListResponse>(url);
         return members;
     }
 
-    public async link(newsroomId: NewsroomId, payload: LinkRequest): Promise<NewsroomRef[]> {
-        const url = routing.newsroomHubUrl.replace(':newsroom_id', String(newsroomId));
-        const { members } = await this.apiClient.post<LinkResponse>(url, { payload });
+    public async linkMany(newsroomId: NewsroomId, payload: LinkManyRequest): Promise<HubMember[]> {
+        const url = generateUrl(newsroomId);
+        const { members } = await this.apiClient.post<ListResponse>(url, { payload });
```
e1himself, line 24: I had to break backward compatibility here.

MohammadxAli (approved):

PR #192, opened 2022-12-27
Feature - Support widget API properties

marko-ogg:


src/types/Newsroom.ts
```diff
@@ -69,6 +69,7 @@ export interface NewsroomRef {
         manual_subscription_management: string;
         privacy_settings: string;
         widget_settings: string;
+        widget_javascript: string;
```
marko-ogg, line 72: @aspirisen this is the widget javascript URL, e.g.: https://bea-theme.prezly.com/widget/javascript

Is that the one you need here?

aspirisen: i am not sure, i do not need the javascript itself. I need the feed layout, like that `https://bea-theme.prezly.com/feed.widget`

marko-ogg: Got it, changed! 👍🏻

aspirisen: @marko-ogg wait a bit, let's discuss in the linear ticket, maybe that would be enough?

src/types/Newsroom.ts
```diff
@@ -69,6 +69,7 @@ export interface NewsroomRef {
         manual_subscription_management: string;
         privacy_settings: string;
         widget_settings: string;
+        widget_feed: string;
```
aspirisen, line 72: @marko-ogg i was able to construct iframe locally, so i think using js widget should be ok

marko-ogg:
This was canceled.

PR #202, opened 2023-01-24
Make cultures property optional in Newsroom create request

kudlajz:


src/endpoints/Newsrooms/types.ts
```diff
@@ -30,7 +30,7 @@ export interface CreateRequest {
      * {subdomain}.prezly.com
      */
     subdomain: string;
-    cultures: CultureRef['code'][];
+    cultures?: CultureRef['code'][];
```
kudlajz, line 33: I'm not sure why it was required before (probably an oversight) as it's possible to create a newsroom without specifying them.

PR #215, opened 2023-02-21
[DEV-10111] Implement `PricingTables` API SDK

MohammadxAli:
Related backend changes: https://github.com/prezly/prezly/pull/12622

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -18,6 +18,7 @@ import {
     NewsroomDomains,
     NewsroomGalleries,
     NewsroomHub,
+    PricingTables,
     SenderAddresses,
     Signup,
     Stories,
@@ -55,6 +56,7 @@ export interface Client {
     newsroomDomains: NewsroomDomains.Client;
     newsroomGalleries: NewsroomGalleries.Client;
     newsroomHub: NewsroomHub.Client;
+    pricingTables: PricingTables.Client;
     senderAddresses: SenderAddresses.Client;
     signup: Signup.Client;
     stories: Stories.Client;
@@ -96,6 +98,7 @@ export function createClient({
         newsroomDomains: new NewsroomDomains.Client(apiClient),
         newsroomGalleries: new NewsroomGalleries.Client(apiClient),
         newsroomHub: new NewsroomHub.Client(apiClient),
+        pricingTables: new PricingTables.Client(apiClient),
         senderAddresses: new SenderAddresses.Client(apiClient),
         signup: new Signup.Client(apiClient),
         stories: new Stories.Client(apiClient),
--- a/src/endpoints/PricingTables/types.ts
+++ b/src/endpoints/PricingTables/types.ts
@@ -2,34 +2,9 @@ export interface PricingTable {
     options: Option[];
 }
 
-export enum BillingCycle {
-    MONTH = 'month',
-    YEAR = 'year',
-}
-
-export enum Currency {
-    EUR = 'eur',
-    USD = 'usd',
-}
-
-export enum OptionId {
-    STARTER = 'starter',
-    CORE = 'core',
-    PREMIUM = 'premium',
-}
-
-export enum LimitDisplayName {
-    STORIES = 'Stories',
-    USERS = 'Users',
-    SITES = 'Sites',
-    CONTACTS = 'Contacts',
-    EMAIL_SENDS = 'Email sends',
-    CUSTOM_SENDER_ADDRESSES = 'Custom sender addresses',
-}
-
 export interface Option {
-    id: `${OptionId}`;
-    display_name: string | null;
+    id: string;
+    display_name: string;
     description: string | null;
     overview_features_list: PlanFeatureRef[];
     limits: Limit[];
@@ -44,16 +19,16 @@ export interface PlanFeatureRef {
 
 export interface Limit {
     id: string;
-    display_name: `${LimitDisplayName}`;
+    display_name: string;
     value: null | number;
     per: null | string;
 }
 
 export interface Price {
-    billing_cycle: `${BillingCycle}`;
-    currency: `${Currency}`;
-    unit_amount: number;
-    unit: string;
+    billing_cycle: string;
+    currency: string;
+    amount: number;
+    unit: string | null;
 }
 
 export interface AddOn {
--- a/src/endpoints/PricingTables/types.ts
+++ b/src/endpoints/PricingTables/types.ts
@@ -1,0 +1,63 @@
+export interface PricingTable {
+    options: Option[];
+}
+
+export enum BillingCycle {
+    MONTH = 'month',
+    YEAR = 'year',
+}
+
+export enum Currency {
+    EUR = 'eur',
+    USD = 'usd',
+}
+
+export enum OptionId {
+    STARTER = 'starter',
+    CORE = 'core',
+    PREMIUM = 'premium',
+}
+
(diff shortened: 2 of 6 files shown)


fgyimah (approved):
🥇

src/endpoints/PricingTables/types.ts
```diff
@@ -0,0 +1,63 @@
...
+    USERS = 'Users',
+    SITES = 'Sites',
+    CONTACTS = 'Contacts',
+    EMAIL_SENDS = 'Email sends',
+    CUSTOM_SENDER_ADDRESSES = 'Custom sender addresses',
+}
+
+export interface Option {
+    id: `${OptionId}`;
+    display_name: string | null;
```
marko-ogg, line 32: It is ok to make this not null.

src/endpoints/PricingTables/types.ts
```diff
@@ -0,0 +1,63 @@
...
+    STORIES = 'Stories',
+    USERS = 'Users',
+    SITES = 'Sites',
+    CONTACTS = 'Contacts',
+    EMAIL_SENDS = 'Email sends',
+    CUSTOM_SENDER_ADDRESSES = 'Custom sender addresses',
+}
+
+export interface Option {
+    id: `${OptionId}`;
```
marko-ogg, line 31: Will this give the SDK problems if we add new pricing option in the API but don't support it in the SDK?

Same question for `Limit.display_name`, `Price.billing_cycle` and `Price.currency`.

src/endpoints/PricingTables/types.ts
```diff
@@ -0,0 +1,63 @@
...
+    display_name: `${LimitDisplayName}`;
+    value: null | number;
+    per: null | string;
+}
+
+export interface Price {
+    billing_cycle: `${BillingCycle}`;
+    currency: `${Currency}`;
+    unit_amount: number;
+    unit: string;
```
marko-ogg, line 56: Unit can be `null`, in case of agency plans for example.

marko-ogg (commented):
The endpoint is still not open, I've prepared a PR for that, which means at the moment it requires authentication.

After I open the endpoint, will it require any changes here in SDK?

src/endpoints/PricingTables/types.ts
marko-ogg: Maybe I brought confusion with `unit_amount` property? Do you think we should just call it `amount`?

MohammadxAli: Yeap I believe `amount` makes more sense here.

src/endpoints/PricingTables/types.ts
marko-ogg: Then I believe it's best to make it `string`, since this way it will complicate the API updates.

marko-ogg: Also, most of these enums already exist, see `Billing/types.ts`. I don't think `LimitDisplayName` should be an enum, since backend should have freedom on those.

src/Client.ts
```diff
@@ -96,6 +98,7 @@ export function createClient({
         newsroomDomains: new NewsroomDomains.Client(apiClient),
         newsroomGalleries: new NewsroomGalleries.Client(apiClient),
         newsroomHub: new NewsroomHub.Client(apiClient),
+        pricingTables: new PricingTables.Client(apiClient),
```
MohammadxAli, line 101: > After I open the endpoint, will it require any changes here in SDK?

Actually, I'm not sure about the best approach here, since `createClient` will require an `accessToken` and it is required:
https://github.com/prezly/javascript-sdk/blob/028170fc03159c90736382fa062618fe96d3cdbe/src/api/ApiClient.ts#L8

- Maybe we should also add a new client creator that doesn't require `accessToken`? Something like `createPublicClient`?
- Or make the `accessToken` optional?

@e1himself @aspirisen @fgyimah what do you suggest here?

fgyimah: We could pass in a fixed access token just as we do for [prezly-to-contentful](https://github.com/prezly/prezly-to-contentful) and that could work, @marko-ogg can give more details if that can help.

See example [here](https://github.com/prezly/prezly-to-contentful/blob/635056039e526b2c19f0df48f2e923305c5da8a5/pages/api/newsrooms.ts#L11)

MohammadxAli: I just realized we have another public API (`register`) that doesn't require an `accessToken`, I looked into app for inspiration and I believe we won't need to change anything here in SDK.

Achievable by overriding `header` when creating client:

```ts
createPrezlyClient({
    accessToken: '',
    headers: { authorization: '' },
});
```

src/endpoints/PricingTables/types.ts
marko-ogg: This is changed in the API.

MohammadxAli: https://github.com/prezly/javascript-sdk/pull/215/commits/dfd2002b381e9b832b6077669f586ea6a27ea78b

src/endpoints/PricingTables/types.ts
MohammadxAli: https://github.com/prezly/javascript-sdk/pull/215/commits/dfd2002b381e9b832b6077669f586ea6a27ea78b

src/endpoints/PricingTables/types.ts
MohammadxAli: I find it really weird that we prefix those enums in `Billing/types.ts` with `Signup`.

https://github.com/prezly/javascript-sdk/blob/dfd2002b381e9b832b6077669f586ea6a27ea78b/src/endpoints/Billing/types.ts#L1-L16

So in order to access them in frontend we do something like:
```tsx
Billing.SignupPlan
Billing.SignupCurrency
Billing.SignupBillingCycle
```

Makes more sense to just do:
```tsx
Billing.Plan
Billing.Currency
Billing.Cycle
```

But I have removed the PricingTable enums in favor of those: https://github.com/prezly/javascript-sdk/pull/215/commits/dfd2002b381e9b832b6077669f586ea6a27ea78b

marko-ogg (approved):

src/endpoints/PricingTables/types.ts
aspirisen: @MohammadxAli I do not quite understand how adding new enum value can cause TS error. Could you explain a bit more what did you mean?

> Yeah, they will pop up a TS error. Do you think I should change them to string instead? I wanted to be extra type-safe here :)

MohammadxAli: Hey @aspirisen, what I meant was:

```ts
enum SomeEnum {
  OLD_VALUE = 'old'
}

SomeEnum.NEW_VALUE // -> Property 'NEW_VALUE' does not exist on type 'typeof SomeEnum'
```
Because it needs to be added to SDK. I didn't mean _adding new enum value on backend side can cause TS error_ 😬

aspirisen: Ah, got what you mean. I think it is fair to update sdk first before using new values in the app, the same we have for objects for example

MohammadxAli: Exactly, I had mixed feelings when deleted enums as well. But I thought we want to give SDK more freedom when we discussed it, but now I realize it was the wrong thing to do.

We always want to update the SDK in the first place.

aspirisen: Agree 👍

PR #222, opened 2023-03-13
Add currency and billing_cycle types

aspirisen:


src/types/Plan.ts
```diff
@@ -40,3 +42,8 @@ export enum PlanLevel {
     AGENCY_LARGE = 'agency_large',
     AGENCY_HUGE = 'agency_huge',
 }
+
+export enum PlanBillingCycle {
```
marko-ogg, line 46: There is a `SignupBillingCycle` already. Not sure if we should just define a general `BillingCycle` type to use in both?

marko-ogg (approved):

PR #223, opened 2023-03-13
[DEV-10250] Feature - Define plan possible changes and ending dates

marko-ogg:


src/types/Plan.ts
```diff
@@ -21,13 +21,28 @@ export interface Plan extends PlanReference {
...
+    ended_at: string | null;
 }
 
 export interface Usage {
     limit: Limit;
     used: number;
 }
 
+export interface Change {
+    pricing_table_option_id: string;
```
aspirisen, line 35: What should i do to get list of plans card?

marko-ogg: It's in the Pricing Tables API. You just list the tables from there, you will get all the pricing options.

aspirisen:
@marko-ogg could you also add Plans enum

aspirisen:
> "pricing_table_option_id": PricingTableOptionId | string;

That will not work as the final type will be just `string` as the widest type and TS will no check that

> I remember we avoided adding the enum as property type because of the possible additions to the API in future (see https://github.com/prezly/javascript-sdk/pull/215#discussion_r1113136254).

As i remember there were no cases where I had to use explicit plan IDs and we just used a set of flags. 
But now we need plan IDs for showing CTAs. I am not sure if it is good that in sdk the field will be just string but in the app code I will "magically" use plan IDs.

When there will be new plans can we just add them in sdk, or they can be random without static definition?

marko-ogg:
> When there will be new plans can we just add them in sdk, or they can be random without static definition?

@aspirisen we can add them to SDK but I don't think there should be such tight dependency between API and SDK. I mean, adding new plans shouldn't be able to break the SDK, no?

What about defining the enum in our app, as a helper to do the checks?

aspirisen:
I am not sure what is the point - API sends us data but we can't define it in sdk?

marko-ogg:
I just don't feel like defining enum on IDs is the right thing to do. It makes sense for types, statuses, currencies, but IDs can be anything we define.

Since you'll be using it, I will let you decide on it. If it's gonna make the work easier, let's add it.

marko-ogg:
@aspirisen I've added the `OptionId` enum, and support for pricing tables list API. Please take a look.

aspirisen (approved):

PR #241, opened 2023-06-02
[DEV-10922] Add support for v4 Stories format

e1himself:
Based on:
- https://github.com/prezly/prezly/pull/13299
- https://github.com/prezly/prezly/pull/13309
- https://github.com/prezly/prezly/pull/13308
- https://github.com/prezly/prezly/pull/13320

### DIFF
--- a/src/endpoints/Stories/Client.ts
+++ b/src/endpoints/Stories/Client.ts
@@ -51,9 +51,7 @@ export class Client {
         StoryRecord extends Story = Options['include'] extends Include
             ? Story & Pick<Story.ExtraFields, Options['include'][number]>
             : Story,
-    >(
-        options?: Exactly<ListOptions<Include>, Options> & { formats?: Formats },
-    ): Promise<ListResponse<StoryRecord>> {
+    >(options?: Exactly<ListOptions<Include>, Options>): Promise<ListResponse<StoryRecord>> {
         const { limit, offset, sortOrder, include, formats } = options ?? {};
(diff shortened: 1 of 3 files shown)


MohammadxAli (approved):

src/endpoints/Stories/Client.ts
```diff
@@ -49,9 +51,12 @@ export class Client {
         StoryRecord extends Story = Options['include'] extends Include
             ? Story & Pick<Story.ExtraFields, Options['include'][number]>
             : Story,
-    >(options?: Exactly<ListOptions<Include>, Options>): Promise<ListResponse<StoryRecord>> {
-        const { limit, offset, sortOrder, include } = options ?? {};
+    >(
+        options?: Exactly<ListOptions<Include>, Options> & { formats?: Formats },
```
oleg-semyonov, line 55: I wonder if `formats` property could be included in the `ListOptions`/`SearchOptions`, etc. types? 🤔 
It looks like this property is added to almost every endpoint, so placing it as a first-party member would be a bit cleaner and feel less like a custom add-on

oleg-semyonov (approved):

src/endpoints/Stories/Client.ts
e1himself: I see what you mean. I have initially done it the way you describe, but then `formats` option declaration was split between two files, which I didn't like. 

Though your point does very much make sense too. I guess it'd be better to revert it back :) 

Thanks!

PR #258, opened 2023-11-06
Make `Culture[code]` typing stricter

e1himself:
To avoid mistakes of misusing ISO code with dashes as Prezly Culture codes.

E.g. `nl_BE` vs `nl-BE`

This problem shines in the `@prezly/theme-kit-intl` package, where different types of codes are being used in different contexts.

I saw this approach in the Next.js codebase: [alternative-urls-types.ts:418-423](https://github.com/vercel/next.js/blob/99e584c70e6277d06c0aca2a8c822f154988adde/packages/next/src/lib/metadata/types/alternative-urls-types.ts#L418-L423)

### DIFF
--- a/src/types/Category.ts
+++ b/src/types/Category.ts
@@ -5,7 +5,7 @@ export interface CategoryRef {
     display_name: string;
     display_description: string | null;
     i18n: {
-        [localeCode: string]: {
+        [localeCode: CultureRef['code']]: {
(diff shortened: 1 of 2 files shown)


fgyimah (approved):
👍🏾

oleg-semyonov (approved):
Nice! 👏🏻

PR #260, opened 2023-11-07
[DEV-11945] Declare `Category.Translation` type

e1himself:


src/types/Category.ts
```diff
@@ -1,21 +1,25 @@
...
     display_description: string | null;
-    i18n: {
-        [localeCode: CultureRef['code']]: {
-            description: string | null;
-            locale: CultureRef;
-            name: string;
-            slug: string | null;
-        };
-    };
+    i18n: Partial<{
```
e1himself, line 7: I've also made it use `Partial<...>` to hint typescript that not every locale code is defined.

Without it, Typescript is always happy seeing `category['nl'].name`, even though the `nl` locale may not be present in the `i18n` object.

PR #265, opened 2023-12-07
[DEV-12120] Implement a new Stories API to fetch story by slug

e1himself:
Implementation: https://github.com/prezly/prezly/pull/14124

The difference with `search()` is that this one can find relocated stories by its former slug values, even after slug has been changed.

The consumer code is expected to perform a redirect when a story slug doesn't match the one coming from the request.


kudlajz (approved):

fgyimah (approved):
👍🏾

PR #267, opened 2023-12-11
[CARE-3589] Upgrade `typescript` to the latest version

e1himself:
- Bump `typescript` to 5.3+
- Bump `node` to v20 (LTS)
- Switch tests from `ts-jest` to `vitest` (and rewrite them to Typescript)
- Drop `package-lock.json` from the repo


.nvmrc
```diff
@@ -1 +1 @@
-v16
\ No newline at end of file
+v20
```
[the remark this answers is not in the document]
oleg-semyonov: It shouldn't, as long as we don't access APIs that are only present in newer versions of Node.

src/http/lib.ts
```diff
@@ -1,26 +1,8 @@
 import type { ParsedQuery } from 'query-string';
 import queryString from 'query-string';
-import * as nodeUrl from 'url';
 
 import type { DeferredJobResponse } from './types';
 
-/**
- * Choose which `URL` class to use based on the environment (Browser / Node.js / Edge)
- */
-function getURLClass() {
```
oleg-semyonov, line 10: Are you sure this won't break things when using SDK in Edge environment? I think when I was adding this code, we were already using Node 16, so it might be that applications that run on Node 16 might break.

.nvmrc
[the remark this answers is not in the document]
kudlajz: I believe it should not affect anything if the package is just imported, as it would be otherwise very dangerous.

src/http/lib.ts
e1himself: > Are you sure this won't break things when using SDK in Edge environment?

Tried running it in the Cloudflare Worker Edge environment. I confirm it works:

Screenshot from 2023-12-11 16-24-43

e1himself: > so it might be that applications that run on Node 16 might break.

True. But wouldn't it be solvable by polyfilling the `URL` API in the specific environment, similar to `fetch`?

oleg-semyonov (approved):

MohammadxAli (approved):

PR #285, opened 2024-03-03
Refactor - Avoid classes & replace them with factory function + POJO

e1himself:
- Avoid classes & replace them with factory function + POJO
- Additionally, unify `remove` and `delete` functions naming to prefer `delete`

## Reasoning

Classes are structurally typed sometimes and nominally typed other times. 
If a class has any private fields, then TypeScript treats it as a nominal type. Otherwise, it's a structural type. 

This has become a problem when I've been trying to use `ContentDelivery` from `@prezly/theme-kit-core` with a different (but compatible) version of `@prezly/sdk` Client object. It complained that `DeferredJobsApiClient` versions are not compatible because of their private fields. :man_shrugging:


MohammadxAli (approved):
Cool! 👍

PR #288, opened 2024-03-13
[DEV-12629] Add `is_featured`, `image` to Category

pestaa:


src/endpoints/NewsroomCategories/types.ts
```diff
@@ -18,10 +20,12 @@ export interface CreateRequest {
             description?: string;
         };
     };
+    image?: UploadedImage | null;
```
kudlajz, line 23: Hey @pestaa, I think this should be a `string | null`? 🤔 I've tried providing the `UploadedImage` directly but

PR #289, opened 2024-04-25
[DEV-12883] Feature - Declare Contact Tags API + extend it

e1himself:
Partially depends on https://github.com/prezly/prezly/pull/14697

### DIFF
--- a/src/endpoints/ContactTags/Client.ts
+++ b/src/endpoints/ContactTags/Client.ts
@@ -1,6 +1,6 @@
 import type { DeferredJobsApiClient } from '../../api';
 import { routing } from '../../routing';
-import type { Contact, ContactTag } from '../../types';
+import { type Contact, type ContactTag, Query } from '../../types';
 import { SortOrder } from '../../types';
 
 import type {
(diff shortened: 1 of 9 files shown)


src/endpoints/ContactTags/Client.ts
```diff
@@ -0,0 +1,138 @@
...
+type RawModificationResponse = {
+    tag: ContactTag;
+    deleted_tags_ids: ContactTag['id'][];
+};
+type RawDeleteResponse = {
+    deleted_tags_ids: ContactTag['id'][];
+};
+
+export function createClient(api: DeferredJobsApiClient) {
+    async function list(options: ListOptions = {}): Promise<ListResponse> {
```
kudlajz, line 33: Is there a reason to have both `list` and `search` since `search` is basically a `list` with a support for `query`? 🤔

kudlajz (approved):

src/endpoints/ContactTags/Client.ts
e1himself: We have this convention in many other endpoints: 

- List API is using a GET request and does not support `query` (because complex query expressions can easily overflow the max URI length)
- Search PI is using a POST request and does support `query` (as part of the request body)

Now the question is whether it is beneficial to have a separate List API at all, and not just replace everything with Search? IMO it is. One quick argument I have from the top of my head -- GET requests can be easily cached, while POST requests usually are not. Are there more pros? :thinking:

kudlajz: Alright 🙂

PR #291, opened 2024-05-06
[CARE-4942] Declare the new `onetrust_cookie_consent` API property

e1himself:
Requires https://github.com/prezly/prezly/pull/14746

### DIFF
--- a/src/types/Newsroom.ts
+++ b/src/types/Newsroom.ts
@@ -97,9 +97,10 @@ export interface Newsroom extends NewsroomRef {
     custom_data_request_link: string | null;
 
     tracking_policy: Newsroom.TrackingPolicy;
-    cookiepro: {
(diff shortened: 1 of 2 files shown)


src/types/Newsroom.ts
```diff
@@ -97,9 +97,10 @@ export interface Newsroom extends NewsroomRef {
     custom_data_request_link: string | null;
 
     tracking_policy: Newsroom.TrackingPolicy;
-    cookiepro: {
```
e1himself, line 100: This will be a breaking change, so a major version bump will be required.

The old property name will still be there for older API clients. Just not declared in the latest interface.

MohammadxAli (approved):

fgyimah (approved):
👍🏾

PR #296, opened 2024-06-12
Define is_demo properties in Campaign, Contact, Newsroom and Story

marko-ogg:


marko-ogg:
@kudlajz I'm not yet sure if it'll be necessary for tags and segments, so just defined the ones here which I'm pretty sure you'll need at the start.

kudlajz (approved):

kudlajz:
I think we need it for tags as well, at least based on the designs in Figma.

marko-ogg:
> I think we need it for tags as well, at least based on the designs in Figma.

Thanks! I didn't see that.

kudlajz (approved):

PR #301, opened 2024-06-27
[DEV-13075] Feature - Implement contact tag groups SDK

marko-ogg:


src/types/ContactTagGroup.ts
```diff
@@ -0,0 +1,7 @@
+export interface ContactTagGroupRef {
+    id: number;
+    name: string;
+    color: string;
+}
+
+export interface ContactTagGroup extends ContactTagGroupRef {}
```
marko-ogg, line 7: It's empty now but we may expand some props here, while living references as is. Not sure if this separation is necessary ATM? 🤔

fgyimah (approved):
Looks good! 👍🏾

PR #306, opened 2024-07-18
Feature - Add site property to NewsroomContact type

kudlajz:
The API returns this property but it is not declared in the type.


kudlajz:
Hmm, any ideas why the build fails all of a sudden? 🤔

marko-ogg (approved):

e1himself (approved):

PR #307, opened 2024-07-24
[DEV-13162] Feature - Add support for `domain` property in organization `Contact`

MohammadxAli:


src/types/Story.ts
```diff
@@ -24,7 +24,7 @@ export interface StoryRef {
      * @see status
      * @see Story.Status
      */
-    lifecycle_status: Story.LifecycleStatus;
+    lifecycle_status: Story.Status;
```
MohammadxAli, line 27: The build on `master` and this branch was failing. I guess the way we used to export it doesn't work in recent versions of TypeScript or something around our build pipeline:

```tsx
// With TS namespace, doesn't work ❌
export namespace Story {
  export import LifecycleStatus = Status;
};

// Without TS namespace, works ✔️
export import LifecycleStatus = Status;
```

This was a deprecated API so just removed it. I believe it's safe, right?

src/types/Story.ts
```diff
@@ -183,12 +183,6 @@ export namespace Story {
         PUBLISHED = 'published',
     }
 
-    /**
-     * @deprecated Please use `Status` instead.
-     * @see Status
-     */
-    export import LifecycleStatus = Status;
```
e1himself, line 190: > The build on master and this branch was failing. I guess the way we used to export it doesn't work in recent versions of TypeScript or something around our build pipeline

Hm... Might be a change in the newer TS versions. 

A few thoughts: 

1. There are more of `export import` places in this codebase:
   image

2. Can you please try to use `export const T = X; export type T = X;` instead? Will it work?

3. If we're going to remove (which I don't mind), it has to be a major release because of the BC break.

src/types/Contact.ts
```diff
@@ -24,6 +24,7 @@ export interface Contact {
     is_demo: boolean;
 
     company_name: string;
+    domain: string;
```
e1himself, line 27: I believe this property is nullable. 

```suggestion
    domain: string | null;
```

src/types/Story.ts
e1himself: > The build on master and this branch was failing. I guess the way we used to export it doesn't work in recent versions of TypeScript or something around our build pipeline

Hm... Might be a change in the newer TS versions. 

A few thoughts: 

1. There are more of `export import` places in this codebase:
   image

2. Can you please try to use `export const T = X; export type T = X;` instead? Will it work?

3. If we're going to remove (which I don't mind), it has to be a major release because of the BC break.

MohammadxAli: > 2. Can you please try to use `export const T = X; export type T = X;` instead? Will it work?

Works! 👍 🎉

src/types/Story.ts
```diff
@@ -187,7 +187,8 @@ export namespace Story {
      * @deprecated Please use `Status` instead.
      * @see Status
      */
-    export import LifecycleStatus = Status;
+    export const LifecycleStatus = Status;
+    export type LifecycleStatus = Status;
```
e1himself, line 191: Great! Now let's do this for other `export import` statements in this repo. :+1:

MohammadxAli: Done: https://github.com/prezly/javascript-sdk/pull/307/commits/12ed5985240b5c3021e14762c6499d8299c26a9a

e1himself (approved):

PR #309, opened 2024-08-01
Add coverages as referenced entities

pestaa:
* https://github.com/prezly/prezly/pull/15092

### DIFF
--- a/src/types/Story.ts
+++ b/src/types/Story.ts
@@ -498,5 +498,4 @@ export interface ExtendedStory
             | 'content'
             | 'attached_gallery_content'
             | 'referenced_entities'
-        > {
-}
+        > {}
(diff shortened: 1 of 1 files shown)


src/types/Story.ts
```diff
@@ -282,6 +283,7 @@ export namespace Story {
 
         referenced_entities: {
             stories: Record<string, OEmbedInfo>;
+            coverages: Record<number, CoverageEntry>;
```
kudlajz, line 286: I believe `coverage` is correct as it's non-countable word: https://www.britannica.com/dictionary/coverage.
Also, did you know that word of the day is [goulash](https://www.britannica.com/dictionary/eb/word-of-the-day)?

kudlajz (approved):
Just make sure to fix the linter issue 🙂

src/types/Story.ts
pestaa: Well, shoot, I noticed this late. Unfortunately, coverages is all over backend, so it made sense to follow the pattern. 

Similarly, I wouldn't hesitate to say goulashes.

kudlajz: What an absolute mad man.

pestaa:
Thanks for pushing this.

PR #310, opened 2024-08-08
[CARE-5862] Add `screenshot` to the Coverage entry definition

e1himself:


src/types/CoverageEntry.ts
```diff
@@ -29,6 +29,7 @@ export interface CoverageEntry {
     note_content_text: string;
     attachment: UploadedFile | UploadedImage | null;
     attachment_oembed: OEmbedInfo | null;
+    screenshot: UploadedFile | null;
```
kudlajz, line 32: What's this property for? Should it be `UploadedImage`?

e1himself: Right, right, right! Thanks!

e1himself: https://github.com/prezly/javascript-sdk/pull/310/commits/69895beab705515295ee28777336ad984862d229

kudlajz (approved):

PR #313, opened 2024-08-23
[DEV-13404] Accept null for newsroom logos create/update request

fgyimah:


src/endpoints/Newsrooms/types.ts
```diff
@@ -37,11 +37,11 @@ export interface CreateRequest {
     /**
      * Uploadcare Image JSON
      */
-    newsroom_logo?: string;
+    newsroom_logo?: string | null;
     /**
      * Uploadcare Image JSON
      */
-    square_logo?: string;
+    square_logo?: string | null;
```
kudlajz, line 44: I don't think the `null` is needed in the `CreateRequest` because if you don't want to provide a logo, you can just omit the property completely.

kudlajz (approved):
Just one small thing.

src/endpoints/Newsrooms/types.ts
fgyimah: https://github.com/prezly/javascript-sdk/pull/313/commits/0e36cc48f816140596b862cf8a9489cb46bc6d43

PR #316, opened 2024-09-05
[DEV-13436] Declare `enrichments` on the `Contact` interface

e1himself:


src/types/Contact.ts
```diff
@@ -108,7 +111,23 @@ export interface Contact {
...
     };
 
-    social: { type: Contact.SocialNetwork; username: string }[];
+    enrichments: {
+        languages: Language['code'][];
+        medium_types: Contact.MediumType[];
+        urls: string[];
+        phone_numbers: Contact.PhoneNumber[];
+        social: { type: Contact.SocialNetwork; username: string }[];
+        periodicity: Contact.Periodicity | null;
```
kudlajz, line 120: Did you change your mind about always using arrays?

kudlajz (approved):

src/types/Contact.ts
e1himself: Yes. After thinking more on it, I've decided that repeating the structure of the contact API object is more beneficial.

PR #319, opened 2024-09-10
[CARE-6160] Support Coverage Integration Runs API

marko-ogg:
Also added new properties to `CoverageIntegration`: `stats_added` and `creator`.


src/Client.ts
```diff
@@ -104,6 +106,7 @@ export function createClient({
         contactsExports: ContactsExports.createClient(api),
         coverage: Coverage.createClient(api),
         coverageIntegrations: CoverageIntegrations.createClient(api),
+        coverageIntegrationRuns: CoverageIntegrationRuns.createClient(api),
```
yuriyyakym, line 109: @marko-ogg 🤔 not sure about creating additional namespace for that. It's still coverage integration, so maybe it's better if it's placed there under a name like `timeline`, `history`, `listRuns` or something like this.

```ts
prezlySdk.coverageIntegrations.listRuns(coverageIntegrationId)
```

Or do we group APIs by response entity type?

src/endpoints/CoverageIntegrationRuns/Client.ts
```diff
@@ -0,0 +1,33 @@
...
+        const { sortOrder, limit, offset } = options;
+        const url = routing.coverageIntegrationRunsUrl.replace(
+            ':integration_id',
+            String(integrationId),
+        );
+        return api.get<ListResponse>(url, {
+            query: {
+                sort: SortOrder.stringify(sortOrder),
+                limit: limit,
+                offset: offset,
```
yuriyyakym, line 25:
```suggestion
                limit,
                offset,
```

yuriyyakym (approved):

src/Client.ts
marko-ogg: Hmm, your suggestion also makes sense.

Reason I've put it in separate namespace because it's a separate API and based on how we were doing similar things so far. For example, see newsroom galleries. By the same logic, newsroom galleries should go under newsrooms namespace `prezlySdk.newsrooms.listGalleries(newsroomId)`, but it's not. It is actually in it's own namespace.

Maybe integration runs are closer to integrations than newsroom galleries are to newsrooms. 🤷🏻‍♂️ 

All in all, I'm ok with both approaches, whatever you prefer and will make more sense to work with frontend side.

yuriyyakym: I would personally keep it in coverage-integration scope, as I don't think it will ever be used outside of it.
Front-end-wise it doesn't matter.
@e1himself, @kudlajz, @MohammadxAli, @fgyimah what do you think?

e1himself: I'd personally keep in the `coverageIntegrations` namespace, as Yuriy suggests.

kudlajz: I also think it makes sense to keep them together 🙂

marko-ogg: Thanks for feedback, I moved it under the same namespace: https://github.com/prezly/javascript-sdk/pull/319/commits/c486d34339a8c109ffdb7afd52dee74e3bd43005

yuriyyakym (approved):
👍

PR #322, opened 2024-09-18
[CARE-6230] Feature - Store integrationRunUuid in coverage entry

yuriyyakym:
@marko-ogg please implement related property on back-end side. Also let me know if the property should be renamed.
I've noticed that we have both `id` and `uuid` props in integration run entry, so I am unsure if I've chosen the correct one for referencing (uuid).

### DIFF
--- a/src/endpoints/Coverage/types.ts
+++ b/src/endpoints/Coverage/types.ts
@@ -9,6 +9,7 @@ import type {
     Story,
     Query,
     SortOrder,
+    CoverageIntegrationRun,
 } from '../../types';
 
 export type Scope = { story: Story['id'] } | null;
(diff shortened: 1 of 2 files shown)


src/endpoints/Coverage/types.ts
```diff
@@ -55,7 +56,9 @@ export interface UpdateRequest {
     page?: number | null;
 }
 
-export interface CreateRequest extends UpdateRequest {}
+export interface CreateRequest extends UpdateRequest {
+    integrationRunUuid?: CoverageIntegrationRun['uuid'] | null;
```
marko-ogg, line 60: I would rename it:

`integrationRunId?: CoverageIntegrationRun['id'] | CoverageIntegrationRun['uuid'] | null;`

marko-ogg (approved):

PR #330, opened 2024-10-15
[DEV-13699] Refactor - Move enums to a namespace to prevent naming conflicts

mohammadxali:
Even though it didn't cause any troubles till now, but it certainly will at some point.

### DIFF
--- a/src/types/CoverageIntegrationRun.ts
+++ b/src/types/CoverageIntegrationRun.ts
@@ -2,22 +2,24 @@ import type { CoverageIntegrationRef } from './CoverageIntegration';
 
 type Iso8601DateTime = string;
 
-export enum Status {
-    DONE = 'done',
-    ERROR = 'error',
-    RUNNING = 'running',
-}
-
 export interface CoverageIntegrationRun {
     id: number;
     uuid: string;
     created_at: Iso8601DateTime;
     updated_at: Iso8601DateTime;
-    status: Status;
+    status: CoverageIntegrationRun.Status;
     /**
      * Following properties will be `null` when status = 'running'
      */
     stats_processed: number | null;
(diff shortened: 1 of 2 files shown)


src/types/CoverageIntegrationRun.ts
```diff
@@ -2,22 +2,24 @@ import type { CoverageIntegrationRef } from './CoverageIntegration';
...
     /**
      * Following properties will be `null` when status = 'running'
      */
     stats_processed: number | null;
     stats_added: number | null;
     integration: CoverageIntegrationRef;
 }
+
+export namespace CoverageIntegrationRun {
+    export enum Status {
```
mohammadxali, line 20: **Before this change:**

The consumer had to import the type from `@prezly/sdk` like this:

```tsx
import { Status } from "@prezly/sdk";

// Usage
Status

// It's not clear what Status refers to. ✖️
// There's a great chance some other entity will have `Status` too. e.g. a story's `Status` ✖️
```

**After this change:**

```tsx
import { CoverageIntegrationRun } from "@prezly/sdk";

// Usage
CoverageIntegrationRun.Status

// It's clear what status refers to. A coverage integration run ✔️
// No need to worry about a naming conflict with another entity exposing status anymore. ✔️

fgyimah (approved):
👍🏾

e1himself (approved):
Awesome! :heart: 

Please note that this is a BC breaking change, so you need to bump the major version number.

src/types/CoverageIntegrationRun.ts
yuriyyakym: Yes, that `Status` worried me many times. Good job refactoring it!

PR #333, opened 2024-10-17
[DEV-13698] Feature - Mark `newsroom` as optional in coverage integration

mohammadxali:


src/endpoints/CoverageIntegrations/types.ts
```diff
@@ -19,6 +19,7 @@ export interface CreateRequest {
 }
 
 export interface UpdateRequest {
+    newsroom?: Newsroom['id'] | Newsroom['uuid'];
```
mohammadxali, line 22: Please let me know if this is correct. Since this is optional I guess we can update it in update request?

pestaa, line 22:
```suggestion
    newsroom?: Newsroom['id'] | Newsroom['uuid'] | null;
```

Yes, can also remove it.

pestaa (approved):
:+1:

PR #352, opened 2025-01-10
[DEV-19095] Newsroom payload change to include privacy and cookie policies

pestaa:


src/types/Newsroom.ts
```diff
@@ -94,9 +94,31 @@ export interface Newsroom extends NewsroomRef {
...
 
+    policies: {
+        privacy_policy:
+            | {
+                  link: string;
+              }
+            | {
+                  content: string;
+                  origin: 'default' | 'custom';
+              };
```
yuriyyakym, line 111: This could be extracted to a for example Policy or PolicySource type

yuriyyakym (approved):

src/types/Newsroom.ts
pestaa: Ideal would be ContentOrigin, but it feels a bit dramatic for a very niche thing. I'll keep it as is for, and we'll address if this area grows.

PR #368, opened 2025-04-07
Feature - Add offset and limit to Pagination

kudlajz:
Noticed that the API returns these properties as part of the `pagination`. Not sure if it's always returned though?

Either way, it's nice because we can use it to determine whether we can load another page when using infinite query.


pestaa (approved):
Not always present, unfortunately.

kudlajz:
Hmm, that's alright. I guess in that case we don't need this and I'll just work with the local `page` and `pageSize` parameters 🤔

PR #389, opened 2025-07-28
[DEV-21181] Feature - Extend `Newsroom` API presentation

e1himself:
App changes: https://github.com/prezly/prezly/pull/17333

- Add `created_at` to Newsroom and NewsroomRef API objects
- Add `domain` to NewsroomRef API object (previously: only on the full Newsroom object)

### DIFF
--- a/src/types/Newsroom.ts
+++ b/src/types/Newsroom.ts
@@ -1,9 +1,8 @@
 import type { UploadedImage } from '@prezly/uploads';
 
+import type { Iso8601DateTime } from './common';
 import type { CultureRef } from './Culture';
 import type { NewroomThemeRef, NewsroomThemePreset } from './NewsroomTheme';
-
-type Iso8601DateTime = string;
 
 export interface NewsroomRef {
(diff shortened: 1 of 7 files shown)


src/types/Newsroom.ts
```diff
@@ -3,6 +3,8 @@ import type { UploadedImage } from '@prezly/uploads';
 import type { CultureRef } from './Culture';
 import type { NewroomThemeRef, NewsroomThemePreset } from './NewsroomTheme';
 
+type Iso8601DateTime = string;
```
kudlajz, line 6: This type is already declared several times throughout the codebase, should we extract it as a global type and re-use it everywhere?

kudlajz (approved):

src/types/Newsroom.ts
e1himself: > This type is already declared several times throughout the codebase, should we extract it as a global type and re-use it everywhere?

Yeah, probably :)

PR #390, opened 2025-07-31
[DEV-21234] Feature - Stories & Campaigns bulk deletion

yuriyyakym:
Pitches are not part of SDK yet, but they should be supported at some point.

### DIFF
--- a/src/endpoints/Campaigns/Client.ts
+++ b/src/endpoints/Campaigns/Client.ts
@@ -56,7 +56,7 @@ export function createClient(api: DeferredJobsApiClient) {
         return api.delete(`${routing.campaignsUrl}/${id}`);
     }
 
-    async function bulkDelete(
+    function bulkDelete(
         payload: BulkDeletePayload,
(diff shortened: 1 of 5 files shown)


src/endpoints/Campaigns/Client.ts
```diff
@@ -56,6 +56,15 @@ export function createClient(api: DeferredJobsApiClient) {
         return api.delete(`${routing.campaignsUrl}/${id}`);
     }
 
+    async function bulkDelete(
+        payload: BulkDeletePayload,
+    ): ProgressPromise<{ records_deleted_number: number }> {
```
marko-ogg, line 61: Is `ProgressPromise` the correct return type here if the API just returns JSON with `records_deleted_number`?

marko-ogg: I can't say what will future requirements bring, I can only say that right now we delete immediately.

Even for non drafts this implementation should stay the same, but in case we notice some performance issues we will need to rethink it.

I don't think we should worry about that now.

PR #391, opened 2025-07-31
[DEV-21199] Templates endpoint

marko-ogg:


src/types/Template.ts
```diff
@@ -0,0 +1,11 @@
+import type { UserRef } from './User';
+
+export interface Template {
+    id: string;
+    title: string;
+    content: string;
+    creator: UserRef | null;
+    last_modifying_user: UserRef | null;
+    created_at: string;
```
kudlajz, line 9: Can you use the `Iso8601DateTime` common type here instead of just `string`? Same for `modified_at`.

kudlajz (approved):

src/types/Template.ts
```diff
@@ -0,0 +1,12 @@
+import type { Iso8601DateTime } from "./common";
```
kudlajz, line 1: You need to run the linter as well (or fix the quotes).

PR #404, opened 2026-04-22
Add optional `hub` field to Newsroom type

JesseWynants:
## Summary

- Adds an optional `hub` field to the `Newsroom` type, mirroring the additive field introduced by [prezly/prezly#18567](https://github.com/prezly/prezly/pull/18567).
- Exposes `hub.parent` and `hub.siblings[]` (each a `NewsroomRef` + `cultures`) so theme/consumer code can render cross-site language pickers on HUB child newsrooms.
- New `Newsroom.HubPeer` interface captures the shape used for parent/siblings.

## Backend shape

- `hub: null` for standalone newsrooms and HUB parents
- `hub: { parent, siblings }` for HUB children

## Release coordination

Field is typed as optional (`hub?:`) on purpose — this lets us merge and release the SDK independently of the backend PR's deploy order without handing consumers a type that lies about runtime shape. Once the backend is live everywhere, we can drop the `?` in a follow-up.

## Test plan

- [x] `tsc --noEmit` passes
- [ ] Reviewer confirms field shape matches the backend PR
- [ ] After backend 18567 deploys, verify SDK consumers can read `newsroom.hub` on a real HUB child

### DIFF
--- a/src/types/Newsroom.ts
+++ b/src/types/Newsroom.ts
@@ -81,6 +81,7 @@ export interface NewsroomRef {
 
 export interface Newsroom extends NewsroomRef {
     is_hub: boolean;
+    hub?: { parent: Newsroom.HubPeer; siblings: Newsroom.HubPeer[] } | null;
     // extended details
     cultures: CultureRef[];
     campaigns_number: number;
@@ -140,6 +141,10 @@ export interface Newsroom extends NewsroomRef {
 }
 
 export namespace Newsroom {
+    export interface HubPeer extends NewsroomRef {
+        cultures: CultureRef[];
+    }
+
     export enum Status {
         ACTIVE = 'active', // i.e. "Live"
         INACTIVE = 'inactive', // i.e. "Not live"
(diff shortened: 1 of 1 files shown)


src/types/Newsroom.ts
```diff
@@ -81,6 +81,7 @@ export interface NewsroomRef {
 
 export interface Newsroom extends NewsroomRef {
     is_hub: boolean;
+    hub?: { parent: Newsroom.HubPeer; siblings: Newsroom.HubPeer[] } | null;
```
e1himself:
```suggestion
    hub: { parent: Newsroom.HubPeer; siblings: Newsroom.HubPeer[] } | null;
```

What what I see in the implementation (https://github.com/prezly/prezly/pull/18567), the `hub` property is always there (cannot be undefined)

src/types/Newsroom.ts
```diff
@@ -140,6 +141,10 @@ export interface Newsroom extends NewsroomRef {
 }
 
 export namespace Newsroom {
+    export interface HubPeer extends NewsroomRef {
+        cultures: CultureRef[];
```
e1himself: The returned array contains extended `CultureRef` objects - with one extra property: `is_default`. Do you think it's worth to document it here? 


```suggestion
        cultures: (CultureRef & { is_default: boolean })[];
```

e1himself (approved):

